PR #2, opened 2018-09-22
Remove tar command

Jyun-Neng:
It is not necessary to generate `handin.tar`.


jserv:
You shall explain the motivation in Git commit message. The change would decouple CMU specific configurations as well.

PR #5, opened 2019-02-21
Add support of Valgrind to check the memory issue

Naetw:
scripts/driver.py will use Valgrind to run each unittest when option
--valgrind is provided. This commit add two members to class Tracer, a
flag for option --valgrind and a string to represent the main command
for subprocess.call().

Notices:

- Remember to patch alarm function used in harness.c since Valgrind will
  inject extra code during the execution which may cause invocation of
  alarm.
  There are several ways of patching.
    - Replace alarm by isnan in executable.
    - Increase the value of variable `time_limit` declared in harness.c.
    - Remove the alarm function in harness.c.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,14 @@ qtest: qtest.c report.c console.c harness.c queue.o
 test: qtest scripts/driver.py
 	scripts/driver.py
 
+valgrind_existence:
+	@which valgrind 2>&1 > /dev/null
+
+valgrind: qtest valgrind_existence
+	cp qtest qtest.patched
+	sed -i "s/alarm/isnan/g" qtest.patched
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -17,6 +17,14 @@ qtest: qtest.c report.c console.c harness.c queue.o
 test: qtest scripts/driver.py
 	scripts/driver.py
 
+valgrind_existence:
+	@which valgrind 2>&1 > /dev/null
+
+valgrind: qtest valgrind_existence
+	cp qtest qtest.patched
```
jserv: Would you like to use `mktemp` to preserve the copy of file `qtest`?

Naetw: Sure. I will add another commit later.

Naetw:
Done! Please review.

PR #6, opened 2019-02-22
Encapsulate strdup function

afcidk:
The strdup function (if used) when implementing queue will now be hooked
to test_strdup.

### DIFF
--- a/harness.c
+++ b/harness.c
@@ -185,6 +185,16 @@ void test_free(void *p)
     allocated_count--;
 }
 
+char *test_strdup(char *s)
+{
+    size_t len = strlen(s) + 1;
+    void *new = test_malloc(len);
+
+    if (new == NULL)
+        return NULL;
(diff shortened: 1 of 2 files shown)


harness.c
```diff
@@ -185,6 +185,16 @@ void test_free(void *p)
     allocated_count--;
 }
 
+char *test_strdup(char *s)
```
jserv: According to Standard C Library, the prototype of [strdup(3)](http://man7.org/linux/man-pages/man3/strdup.3.html) is as following:
```c
char *strdup(const char *s1);
```

The qualifier `const` should be added.

harness.h
```diff
@@ -10,6 +10,7 @@
 
 void *test_malloc(size_t size);
 void test_free(void *p);
+char *test_strdup(char *s);
```
jserv: Ditto. Change the prototype.

PR #7, opened 2019-02-23
Check entire buffer in check overflow

cjwind:
do_remove_head() checked removes[string_length + 1] only, so
q_remove_head() would pass the test when it try to write data to other
place in the buffer, such as index string_length + 2. Check entire
buffer to fix it.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -300,15 +300,11 @@ bool do_remove_head(int argc, char *argv[])
           Validate memory by checking removes's padding positions are still
           initial value 'X'.
         */
-        bool check_overflow = true;
-        for (int i = string_length + 1; i < string_length + STRINGPAD; i++) {
-            if (removes[i] != 'X') {
-                check_overflow = false;
-                break;
-            }
+        int i = string_length + 1;
+        while (i < string_length + STRINGPAD && removes[i] == 'X') {
+            i++;
         }
-
-        if (!check_overflow) {
+        if (i != string_length + STRINGPAD) {
             report(1,
                    "ERROR: copying of string in remove_head overflowed "
                    "destination buffer.");
--- a/qtest.c
+++ b/qtest.c
@@ -297,8 +297,7 @@ bool do_remove_head(int argc, char *argv[])
         }
 
         bool check_overflow = true;
-        int i = 0;
-        for (i = string_length + 1; i < string_length + STRINGPAD; i++) {
+        for (int i = string_length + 1; i < string_length + STRINGPAD; i++) {
             if (removes[i] != 'X') {
                 check_overflow = false;
                 break;
--- a/qtest.c
+++ b/qtest.c
@@ -296,6 +296,10 @@ bool do_remove_head(int argc, char *argv[])
             ok = false;
         }
 
+        /*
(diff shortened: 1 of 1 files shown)


qtest.c
```diff
@@ -294,7 +294,18 @@ bool do_remove_head(int argc, char *argv[])
         if (removes[0] == '\0') {
             report(1, "ERROR: Failed to store removed value");
             ok = false;
-        } else if (removes[string_length + 1] != 'X') {
+        }
+
+        bool check_overflow = true;
+        int i = 0;
```
jserv: Move the scope of `int i` into for-loop. So, it would look like the following:
```C
for (int i = string_length + 1; i < string_length + STRINGPAD; i++) {
    ...
```

qtest.c
```diff
@@ -294,7 +294,18 @@ bool do_remove_head(int argc, char *argv[])
...
+        for (i = string_length + 1; i < string_length + STRINGPAD; i++) {
```
jserv: Add the algorithm descriptions about checking and validating memory here.

qtest.c
```diff
@@ -294,7 +294,21 @@ bool do_remove_head(int argc, char *argv[])
         if (removes[0] == '\0') {
             report(1, "ERROR: Failed to store removed value");
             ok = false;
-        } else if (removes[string_length + 1] != 'X') {
+        }
+
+        /*
```
jserv: Tweak the indention for comments.

Something like:
```C
/*
  How large is a queue before it's considered big.
  This affects how it gets printed
  and whether cautious mode is used when freeing the list
*/
```

qtest.c
```diff
@@ -294,7 +294,21 @@ bool do_remove_head(int argc, char *argv[])
...
+        Validate memory by checking removes's padding positions are still
+        initial value 'X'.
+        */
+        bool check_overflow = true;
```
jserv: I wonder the necessity of variable `check_overflow`. Can you eliminate it by replacing `for` with `while`?

qtest.c
```diff
@@ -297,18 +297,14 @@ bool do_remove_head(int argc, char *argv[])
...
+          initial value 'X'.
         */
-        bool check_overflow = true;
-        for (int i = string_length + 1; i < string_length + STRINGPAD; i++) {
-            if (removes[i] != 'X') {
-                check_overflow = false;
-                break;
-            }
+        int i = string_length + 1;
+        while (i < string_length + STRINGPAD && removes[i] == 'X') {
```
jserv: Add parentheses. So, it becomes `while ((i < string_length + STRINGPAD) && removes[i] == 'X') {`

qtest.c
```diff
@@ -294,7 +294,17 @@ bool do_remove_head(int argc, char *argv[])
         if (removes[0] == '\0') {
             report(1, "ERROR: Failed to store removed value");
             ok = false;
-        } else if (removes[string_length + 1] != 'X') {
+        }
+
+        /*
+          Validate memory by checking removes's padding positions are still
```
jserv: Check for grammar within the comment.

PR #8, opened 2019-02-23
Print out debug hint for each unformatted file

johnnylord:
The original if block should move into the for loop block

### DIFF
--- a/scripts/pre-commit.hook
+++ b/scripts/pre-commit.hook
@@ -33,14 +33,12 @@ for FILE in $FILES; do
         echo "[!] $FILE does not follow the consistent coding style." >&2
         RETURN=1
     fi
-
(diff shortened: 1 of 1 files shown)


scripts/pre-commit.hook
```diff
@@ -33,14 +33,15 @@ for FILE in $FILES; do
...
-    echo "    clang-format -i $FILE" >&2
-    echo
-fi
+    if [ $RETURN -eq 1 ]; then
+        echo "" >&2
+        echo "Make sure you indent as the following:" >&2
+        echo "    clang-format -i $FILE" >&2
+        echo
+    fi
+
```
jserv: Strip the blank line here.

johnnylord: Done~

PR #9, opened 2019-02-28
Prevent test_free treat free(NULL) as error

tfliao:
In man page of free, "If ptr is NULL, no operation is performed."
We should not treat free(NULL) as error, thus removing marking error,
  just align the behavior of native free, perform no operation.

### DIFF
--- a/harness.c
+++ b/harness.c
@@ -154,8 +154,6 @@ void test_free(void *p)
         return;
     }
     if (p == NULL) {
-        report(MSG_ERROR, "Attempt to free NULL");
-        error_occurred = true;
         return;
     }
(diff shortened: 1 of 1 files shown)


jserv:
While referring to function `free`, you shall explicitly mention [free(3)](https://linux.die.net/man/3/free) because there is another "free" behaving as a command line tool.

jserv:
Use `git commit --amend` to rework your Git commit(s).

tfliao:
Luckily, `git push -f ` doesn't mess up github.
Now commit message updated.

jserv:
Use "prevent ... from ..." rather than "prevent" in the subject of commit messages.

PR #13, opened 2020-01-26
Fix(trace): fix expected values and unknown address access

YuShuanHsieh:
## Why
1. The expected values should be changed after the sorting operation.
2. `do_sort` will access unknown memory address if the queue is null.

## How
1. Fix the expected values of trace-4, trace-5.
2. always return true if it is a null queue.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -485,16 +485,18 @@ bool do_sort(int argc, char *argv[])
     exception_cancel();
     set_noallocate_mode(false);
 
-    list_ele_t *e = q->head;
     bool ok = true;
-    while (ok && e && --cnt) {
-        /* Ensure each element in ascending order */
-        /* FIXME: add an option to specify sorting order */
-        if (strcmp(e->value, e->next->value) > 0) {
-            report(1, "ERROR: Not sorted in ascending order");
-            ok = false;
+    if (q) {
+        list_ele_t *e = q->head;
+        while (ok && e && --cnt) {
+            /* Ensure each element in ascending order */
+            /* FIXME: add an option to specify sorting order */
+            if (strcmp(e->value, e->next->value) > 0) {
+                report(1, "ERROR: Not sorted in ascending order");
+                ok = false;
+            }
+            e = e->next;
         }
-        e = e->next;
     }
 
     show_queue(3);
--- a/traces/trace-04-ops.cmd
+++ b/traces/trace-04-ops.cmd
@@ -11,7 +11,7 @@ it bear
 it gerbil
 size
 sort
-rh dolphin
+rh bear
 rh
 rh
 rh
--- a/traces/trace-05-ops.cmd
+++ b/traces/trace-05-ops.cmd
@@ -15,10 +15,10 @@ rh dolphin
 reverse
 size
 sort
-rh gerbil
 rh bear
+rh bear
+rh gerbil
+rh gerbil
 rh meerkat
-rh gerbil
-rh bear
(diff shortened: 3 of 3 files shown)


qtest.c
```diff
@@ -466,8 +466,8 @@ static bool do_size(int argc, char *argv[])
 
 bool do_sort(int argc, char *argv[])
 {
-    if (argc != 1) {
-        report(1, "%s takes no arguments", argv[0]);
+    if (argc != 1 && !(argc == 2 && strcmp(argv[1], "NULL") == 0)) {
+        report(1, "%s takes 0 or `NULL` value argument", argv[0]);
```
jserv: Don't put "`" character here. Simply use quotation mark.

qtest.c
```diff
@@ -466,8 +466,8 @@ static bool do_size(int argc, char *argv[])
 
 bool do_sort(int argc, char *argv[])
 {
-    if (argc != 1) {
-        report(1, "%s takes no arguments", argv[0]);
+    if (argc != 1 && !(argc == 2 && strcmp(argv[1], "NULL") == 0)) {
```
jserv: Rewrite `strcmp(argv[1], "NULL") == 0)` into `!strcmp(argv[1], "NULL")` for consistency.

traces/trace-07-robust.cmd
```diff
@@ -7,4 +7,4 @@ it dolphin
 rh
 reverse
 size
-sort
+sort NULL
```
jserv: It is a bit weird to me regarding the use of `sort NULL`. In my understanding, `sort` should be as generic as possible. It can accept almost anything including NULL. I don't think it is vital to introduce extra argument for sort since `sort` always perform the operations within queue. What is your consideration?

YuShuanHsieh: You are right.
The reason why I choose to add a new argument is to minimize the effect of other test cases. I think it may always return true if it is a null queue, but I am not really sure about this idea because honestly I am not quite familiar with all test cases.

qtest.c
```diff
@@ -485,6 +485,8 @@ bool do_sort(int argc, char *argv[])
     exception_cancel();
     set_noallocate_mode(false);
 
+    if (q == NULL)
```
jserv: Don't place check here. Instead, let's move to Line 477:
```c
if (q == NULL)
    report(3, "Warning: Calling sort on null queue");
error_check();
```

The function `error_check()` would collect test information and interact with users.

YuShuanHsieh: Sorry, I’m not sure what you mean. 
Cause in my point of view ` error_check() `draw a scope of tests. like
```c
error_check(); //reset error_occurred
// test case
return ok && !error_check();
```
So I still need to handle the null queue after `q_sort(q)` to prevent `q->head` bad access.
Could you give me more about your opinion? Thx :)

jserv: You can follow the similar flow in function `do_size` and `do_reverse`, that just report when NULL pointer is inputed. `qtest` itself does not need to take care about fatal condition. Instead, `qtest` only reports to users and each trace file denotes the criteria of test scenario. In this case, `do_sort` attempts to report the checkpoints without managing to recover.

Consequently, the implementation behind queue would have to couple with the above though.

YuShuanHsieh: I probably catch your point. Please review it again. Thx

jserv:
Thank @YuShuanHsieh for contributing to this project!

PR #14, opened 2020-01-28
Fix(qtest): report an error when strcmp is not equal zero

YuShuanHsieh:


jserv:
It was my bad. Thanks for pointing this.

PR #15, opened 2020-01-28
Feature: add random insert cmd and a test case for sorting performance

YuShuanHsieh:
## Why
Test the worst case of sorting. 

## How
1. Add a `randi` cmd to insert the large number of words with DESC order.
2. Add the `trace-16-perf.cmd `  test file for sorting test.

## Todo
maxScores should be changed. Cause the total score now is 107 haha.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -9,7 +9,9 @@
 #include <strings.h> /* strcasecmp */
 #include <sys/stat.h>
 #include <sys/wait.h>
+#include <time.h>
 #include <unistd.h>
+
 
 /* Our program needs to use regular malloc/free */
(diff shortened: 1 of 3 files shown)


jserv:
Let's split into two pull requests:
1. Provide special parameter with `it` and `it` command to generate random strings. The expected usage might be `if RAND 1000000`. The `traces/trace-06-string.cmd` can be changed accordingly to validate random strings.
2. Append new trace files to reflect sorting performance over linked list. Reference: https://arxiv.org/pdf/1310.7890.pdf

YuShuanHsieh:
No problem. I will read these post first! 😄

PR #16, opened 2020-01-28
Feature(qtest): Add the rand option to generate a random string

YuShuanHsieh:


qtest.c
```diff
@@ -222,8 +247,9 @@ static bool do_insert_head(int argc, char *argv[])
 
 static bool do_insert_tail(int argc, char *argv[])
 {
+    char random_str[MAX_RANDSTR_LEN];
```
jserv: Rename it to `randstr_buf`.

YuShuanHsieh: done.

jserv: You don't have to report "done". Use the button "Resolve conversation" on GitHub instead.

qtest.c
```diff
@@ -156,11 +161,24 @@ static bool do_free(int argc, char *argv[])
     return ok && !error_check();
 }
 
+static void random_string(char *buf, size_t buf_size)
+{
+    size_t len = 0, n;
+    while (len < MIN_RANDSTR_LEN)
+        len = rand() % buf_size;
+
+    for (n = 0; n < len; n++) {
```
jserv: Reduce the scope of `size_t n`. That is, remove its declaration in line 166 and rewrite line 170 as following:
```c
for (size_t n = 0; n < len; n++)
    buf[n] = charset[rand() % (sizeof charset - 1)];
```

qtest.c
```diff
@@ -156,11 +161,24 @@ static bool do_free(int argc, char *argv[])
...
+    buf[len] = '\0';
+}
+
 static bool do_insert_head(int argc, char *argv[])
 {
     char *lasts = NULL;
+    char randstr_buf[MAX_RANDSTR_LEN];
     int reps = 1;
-    bool ok = true;
+    bool ok = true, is_random = false;
```
jserv: Change `is_random` to `need_rand`.

qtest.c
```diff
@@ -156,11 +161,24 @@ static bool do_free(int argc, char *argv[])
     return ok && !error_check();
 }
 
+static void random_string(char *buf, size_t buf_size)
```
jserv: Rename the function to `fill_rand_string`.

qtest.c
```diff
@@ -174,12 +192,19 @@ static bool do_insert_head(int argc, char *argv[])
...
+    }
+
     if (!q)
         report(3, "Warning: Calling insert head on null queue");
     error_check();
 
     if (exception_setup(true)) {
         for (int r = 0; ok && r < reps; r++) {
+            if (is_random)
+                random_string(randstr_buf, MAX_RANDSTR_LEN);
```
jserv: The the second parameter to `sizeof(randstr_buf)`.

jserv:
The subject of git commit messages was inaccurate. It can be "Recognize 'RAND' parameter for insertions."

qtest.c
```diff
@@ -76,11 +81,11 @@ static void console_init()
     add_cmd("new", do_new, "                | Create new queue");
     add_cmd("free", do_free, "                | Delete queue");
     add_cmd("ih", do_insert_head,
-            " str [n]        | Insert string str at head of queue n times "
-            "(default: n == 1)");
+            " str [n]        | Insert string str at head of queue n times. "
+            "Insert a random string if str is RAND (default: n == 1)");
```
jserv: Change the message to "Generate random string(s) if str == RAND".

qtest.c
```diff
@@ -156,11 +161,24 @@ static bool do_free(int argc, char *argv[])
     return ok && !error_check();
 }
 
+static void fill_rand_string(char *buf, size_t buf_size)
+{
+    size_t len = 0;
+    while (len < MIN_RANDSTR_LEN)
```
jserv: What if `buf_size` < `MIN_RANDSTR_LEN`?

YuShuanHsieh: That is a good question. But I think the function is for internal usage and both `buf_size` and minimum length are defined by `MIN_RANDSTR_LEN` and `MAX_RANDSTR_LEN`. So It may reduce the chance that maintainer set wrong numbers to cause segmentation fault. My question is do I still need to add a logic to prevent error occurred for these kind of functions?

jserv: You can place `FIXME` or `TODO` comments for the above considerations.

YuShuanHsieh: I decided to place `TODO` comments. And could you share your experiences about the question? Cause in my experience, I only make sure that the internal function works without much error checks. I want to know how do experienced programmers think. Thanks 😄

jserv: I am thinking of the feasibility to re-use `fill_rand_string` to perform fuzzing: https://en.wikipedia.org/wiki/Fuzzing
Typically speaking, fuzzer might expect shorter strings to validate interpreter and/or qtest internals.

PR #19, opened 2020-01-31
Feature(): Add the trace-16 file for sorting performance test

YuShuanHsieh:
Test sorting performance with random and descending order data.
- 10000: all correct sorting algorithms are expected pass.
- 50000: sorting algorithms with O(n^2) time complexity are expected failed.
- 100000: sorting algorithms with O(nlogn) time complexity are expected pass.


jserv:
The the criteria is perfect to me! Thanks for doing this. Can you mention the information about time complexity in file `trace-16-perf.cmd` as well?

PR #21, opened 2020-02-02
Integrate dudect to analyze O(1) time complexity

afcidk:
Dudect is a tool to deterimine whether a piece of code runs in constant
time or not by given different inputs and measure the execution time.

From the lab requirements, `q_insert_tail` and `q_size` are required to
meet O(1) time complexity. To test if the function runs in constant
time, simply open the option with `option simulation 1`, then run either
`it` or `size`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 CC = gcc
-CFLAGS = -O1 -g -Wall -Werror
+CFLAGS = -O1 -g -Wall -Werror -Idudect -I.
 
 GIT_HOOKS := .git/hooks/applied
+DUT_DIR := dudect
 all: $(GIT_HOOKS) qtest
 
 # Control the build verbosity
@@ -24,14 +25,16 @@ $(GIT_HOOKS):
 	@scripts/install-git-hooks
 	@echo
 
-OBJS := qtest.o report.o console.o harness.o queue.o
+OBJS := qtest.o report.o console.o harness.o queue.o \
+        random.o dudect/constant.o dudect/fixture.o dudect/ttest.o
 deps := $(OBJS:%.o=.%.o.d)
 
 qtest: $(OBJS)
 	$(VECHO) "  LD\t$@\n"
-	$(Q)$(CC) $(LDFLAGS)  -o $@ $^
+	$(Q)$(CC) $(LDFLAGS) -o $@ $^ -lm
 
 %.o: %.c
+	@mkdir -p .$(DUT_DIR)
 	$(VECHO) "  CC\t$@\n"
 	$(Q)$(CC) -o $@ $(CFLAGS) -c -MMD -MF .$@.d $<
 
@@ -58,6 +61,7 @@ valgrind: valgrind_existence
 
 clean:
 	rm -f $(OBJS) $(deps) *~ qtest /tmp/qtest.*
+	rm -rf .$(DUT_DIR)
 	rm -rf *.dSYM
 	(cd traces; rm -f *~)
 
--- a/console.c
+++ b/console.c
@@ -17,6 +17,7 @@
 #include "report.h"
 
 /* Some global values */
+int simulation = 0;
 static cmd_ptr cmd_list = NULL;
 static param_ptr param_list = NULL;
 static bool block_flag = false;
@@ -54,7 +55,7 @@ static int fd_max = 0;
 /* Parameters */
 static int err_limit = 5;
 static int err_cnt = 0;
-static int echo = 0;
+static bool echo = 0;
 
 static bool quit_flag = false;
 static char *prompt = "cmd> ";
@@ -98,6 +99,7 @@ void init_cmd()
     add_cmd("log", do_log_cmd, " file           | Copy output to file");
     add_cmd("time", do_time_cmd, " cmd arg ...    | Time command execution");
     add_cmd("#", do_comment_cmd, " ...            | Display comment");
+    add_param("simulation", &simulation, "Start/Stop simulation mode", NULL);
     add_param("verbose", &verblevel, "Verbosity level", NULL);
     add_param("error", &err_limit, "Number of errors until exit", NULL);
     add_param("echo", &echo, "Do/don't echo commands", NULL);
--- a/dudect/constant.c
+++ b/dudect/constant.c
@@ -1,0 +1,84 @@
+#include "constant.h"
+#include <assert.h>
+#include <signal.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include "cpucycles.h"
+#include "queue.h"
+#include "random.h"
+
+/* Allow random number range from 0 to 65535 */
+const size_t chunk_size = 16;
+/* Number of measurements per test */
+const size_t number_measurements = 150;
+const int drop_size = 20;
+/* Maintain a queue independent from the qtest since
+ * we do not want the test to affect the original functionality
+ */
+static queue_t *q = NULL;
+static char random_string[100][8];
+static int random_string_iter = 0;
+enum { test_insert_tail, test_size };
+
+/* Implement the necessary queue interface to simulation */
+void init_dut(void)
+{
+    q = NULL;
+}
+
+char *get_random_string(void)
+{
+    random_string_iter = (random_string_iter + 1) % number_measurements;
+    return random_string[random_string_iter];
+}
+
+void prepare_inputs(uint8_t *input_data, uint8_t *classes)
+{
+    randombytes(input_data, number_measurements * chunk_size);
+    for (size_t i = 0; i < number_measurements; i++) {
+        classes[i] = randombit();
+        if (classes[i] == 0)
+            *(uint16_t *) (input_data + i * chunk_size) = 0x00;
+    }
+
+    for (size_t i = 0; i < 100; ++i) {
+        /* Generate random string */
+        randombytes((uint8_t *) random_string[i], 7);
+        random_string[i][7] = 0;
+    }
+}
+
+void measure(int64_t *before_ticks,
+             int64_t *after_ticks,
+             uint8_t *input_data,
+             int mode)
+{
+    assert(mode == test_insert_tail || mode == test_size);
+    if (mode == test_insert_tail) {
+        for (size_t i = drop_size; i < number_measurements - drop_size; i++) {
+            char *s = get_random_string();
+            dut_new();
+            dut_insert_head(
+                get_random_string(),
+                *(uint16_t *) (input_data + i * chunk_size) % 10000);
+            before_ticks[i] = cpucycles();
+            dut_insert_tail(s, 1);
+            after_ticks[i] = cpucycles();
+            dut_free();
+        }
+    } else {
+        for (size_t i = drop_size; i < number_measurements - drop_size; i++) {
+            dut_new();
+            dut_insert_head(
+                get_random_string(),
+                *(uint16_t *) (input_data + i * chunk_size) % 10000);
+            before_ticks[i] = cpucycles();
+            dut_size(1);
+            after_ticks[i] = cpucycles();
+            dut_free();
+        }
+    }
+}
--- a/dudect/cpucycles.h
+++ b/dudect/cpucycles.h
@@ -1,0 +1,12 @@
+#include <stdint.h>
+// http://www.intel.com/content/www/us/en/embedded/training/ia-32-ia-64-benchmark-code-execution-paper.html
+inline int64_t cpucycles(void)
+{
+#if defined(__i386__) || defined(__x86_64__)
+    unsigned int hi, lo;
+    __asm__ volatile("rdtsc\n\t" : "=a"(lo), "=d"(hi));
+    return ((int64_t) lo) | (((int64_t) hi) << 32);
+#else
+#error Unsupported Architecture
+#endif
+}
--- a/dudect/fixture.c
+++ b/dudect/fixture.c
@@ -1,0 +1,195 @@
+/** dude, is my code constant time?
+ *
+ * This file measures the execution time of a given function many times with
+ * different inputs and performs a Welch's t-test to determine if the function
+ * runs in constant time or not. This is essentially leakage detection, and
+ * not a timing attack.
+ *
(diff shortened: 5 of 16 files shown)


jserv:
I would like to invite @YuShuanHsieh for reviewing this pull request.

Makefile
```diff
@@ -14,7 +15,7 @@ else
 endif
 
 # Enable sanitizer(s) or not
-ifeq ("$(SANITIER)","1")
+ifeq ("$(SANITIZER)","1")
```
jserv: Send another pull request to fix typo.

dudect/fixture.c
```diff
@@ -0,0 +1,316 @@
...
+#include "fixture.h"
+#include <math.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "../console.h"
+#include "constant.h"
+#include "cpucycles.h"
+#include "percentile.h"
```
jserv: File `dudect/percentile.h` is included only in `dudect/fixture.c`. How about moving the lines into file `dudect/fixture.c`?

scripts/driver.py
```diff
@@ -15,47 +16,52 @@ class Tracer:
...
+        8: "trace-08-robust",
+        9: "trace-09-robust",
+        10: "trace-10-malloc",
+        11: "trace-11-malloc",
+        12: "trace-12-malloc",
+        13: "trace-13-perf",
+        14: "trace-14-perf",
+        15: "trace-15-perf",
+        16: "trace-16-perf",
+        17: "trace-17-const"
```
jserv: Rename from `const` to `complexity`.

console.c
```diff
@@ -17,6 +17,7 @@
 #include "report.h"
 
 /* Some global values */
+int simulation = 0;
```
jserv: Should `simulation` be of bool type?

dudect/cpucycles.c
```diff
@@ -0,0 +1,11 @@
+#include "cpucycles.h"
+#include <stdint.h>
+#include <time.h>
+
```
jserv: Add build time check for x86 or x86_64 architecture.

dudect/cpucycles.c
```diff
@@ -0,0 +1,11 @@
...
+// http://www.intel.com/content/www/us/en/embedded/training/ia-32-ia-64-benchmark-code-execution-paper.html
+int64_t cpucycles(void)
```
jserv: Move the implementation of `cpucycles` into header `dudect/cpucycles.h` as a static inline function.

dudect/random.h
```diff
@@ -0,0 +1,9 @@
+#ifndef DUDECT_RANDOM_H
+#define DUDECT_RANDOM_H
+
+#include <stddef.h>
+#include <stdint.h>
+void randombytes(uint8_t *x, size_t xlen);
```
jserv: How about moving file `dudect/random.h` to top-level? Thus, original `qtest.c` can share the functionality as well.

dudect/ttest.c
```diff
@@ -0,0 +1,69 @@
...
+{
+    for (int class = 0; class < 2; class ++) {
+        ctx->mean[class] = 0.0;
+        ctx->m2[class] = 0.0;
+        ctx->n[class] = 0.0;
+    }
+    return;
+}
+
+#if 0
```
jserv: Do we have to leave the dead code?

dudect/fixture.c
```diff
@@ -0,0 +1,316 @@
...
+        // measurements).
+        // Centered product pre-processing.
+        if (t[0]->n[0] > 10000) {
+            double centered = (double) difference - t[0]->mean[classes[i]];
+            t_push(t[1 + number_percentiles], centered * centered, classes[i]);
+        }
+    }
+}
+
+#if 0
```
jserv: Ditto. Dead code.

dudect/percentile.c
```diff
@@ -0,0 +1,49 @@
...
+int64_t percentile(int64_t *a, double which, size_t size)
+{
+    qsort(a, size, sizeof(int64_t), (int (*)(const void *, const void *)) cmp);
+    size_t array_position = (size_t)((double) size * (double) which);
+    assert(array_position < size);
+    // printf("percentile %f is %lld\n", which, a[array_position]);
+    return a[array_position];
+}
+
+#if 0
```
jserv: Ditto. Dead code.

console.c
afcidk: I think`simulation` should be of bool type, but then I saw [`echo`](https://github.com/sysprog21/lab0-c/blob/b9ed05dd64faf00851d73843dddc04e9de3034c0/console.c#L57), which is declared as an integer. Should `echo` also be declared as integer type?

dudect/cpucycles.c
jserv: Any alternative way for measurement in portable fashion?

console.c
jserv: I think echo should be declared as `bool`. The original code taken from CMU ICS was not consistent though. Let's fix.

dudect/constant.c
```diff
@@ -0,0 +1,65 @@
...
+int test_mode;
+/* Maintain a queue independent from the qtest */
+static queue_t *q = NULL;
+
+/* Implement the necessary queue interface to simulation */
+void dut_new()
+{
+    q = q_new();
+}
+
```
jserv: Since functions like `dut_size` and `dut_insert_head` are built with the identical logic, can we use macro to provide these implementations?

dudect/fixture.c
```diff
@@ -0,0 +1,316 @@
...
+
+static void measure(int64_t *before_ticks,
+                    int64_t *after_ticks,
+                    uint8_t *input_data)
+{
+    if (test_mode == test_insert_tail) {
+        for (size_t i = drop_size; i < number_measurements - drop_size; i++) {
+            dut_new();
+            dut_insert_head(
+                "abc", *(uint16_t *) (input_data + i * chunk_size) % 10000);
```
jserv: Why `abc` here?

afcidk: What the string is should not affect the complexity, so I gave a pretty aribitrary string just to insert into the linked list. I don't think that might affect the experiment, or should I change it to random string everytime?

jserv: You can rotate the predefined string list to specify one to be inserted.

dudect/constant.c
```diff
@@ -0,0 +1,93 @@
...
+            dut_new();
+            dut_insert_head(
+                get_random_string(),
+                *(uint16_t *) (input_data + i * chunk_size) % 10000);
+
+            char *s = get_random_string();
+            before_ticks[i] = cpucycles();
+            dut_insert_tail(s, 1);
+            after_ticks[i] = cpucycles();
+            // printf("size: %d, time: %ld\n", *(uint16_t *) (input_data + i *
```
jserv: Remove dead code.

dudect/constant.c
```diff
@@ -0,0 +1,93 @@
...
+        for (size_t i = drop_size; i < number_measurements - drop_size; i++) {
+            dut_new();
+            dut_insert_head(
+                get_random_string(),
+                *(uint16_t *) (input_data + i * chunk_size) % 10000);
+
+            before_ticks[i] = cpucycles();
+            dut_size(1);
+            after_ticks[i] = cpucycles();
+            // printf("size: %d, time: %ld\n", *(uint16_t *) (input_data + i *
```
jserv: Ditto. Dead code.

dudect/cpucycles.h
```diff
@@ -0,0 +1,12 @@
+#include <stdint.h>
+// http://www.intel.com/content/www/us/en/embedded/training/ia-32-ia-64-benchmark-code-execution-paper.html
+inline int64_t cpucycles(void)
+{
+#if defined(__i386__) || defined(__x86_64__)
+    unsigned int hi, lo;
+    __asm__ volatile("rdtsc\n\t" : "=a"(lo), "=d"(hi));
+    return ((int64_t) lo) | (((int64_t) hi) << 32);
+#else
+    Unsupported Architecture
```
jserv: Use `#error`

qtest.c
```diff
@@ -250,6 +251,20 @@ static bool do_insert_head(int argc, char *argv[])
 
 static bool do_insert_tail(int argc, char *argv[])
 {
+    if (simulation) {
+        if (argc != 1) {
+            report(1, "%s does not need arguments in simulation mode", argv[0]);
+            return false;
+        } else {
+            bool ok = is_insert_tail_const();
```
jserv: Use `}` rather than `} else {`. No more `else` is required.

Makefile
```diff
@@ -24,14 +25,16 @@ $(GIT_HOOKS):
 	@scripts/install-git-hooks
 	@echo
 
-OBJS := qtest.o report.o console.o harness.o queue.o
+OBJS := qtest.o report.o console.o harness.o queue.o \
+		random.o dudect/constant.o dudect/fixture.o dudect/ttest.o
```
jserv: Indent with spaces.

console.c
```diff
@@ -17,6 +17,7 @@
 #include "report.h"
 
 /* Some global values */
+bool simulation = 0;
```
jserv: shall be `false`.

console.c
```diff
@@ -54,7 +55,7 @@ static int fd_max = 0;
 /* Parameters */
 static int err_limit = 5;
 static int err_cnt = 0;
-static int echo = 0;
+static bool echo = 0;
```
jserv: Ditto. shall be `false`.

dudect/constant.c
```diff
@@ -0,0 +1,85 @@
+#include "constant.h"
+#include <assert.h>
+#include <signal.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include "../queue.h"
```
jserv: Why did you use `../` here? Can you make use of CFLAGS for search path?

dudect/constant.c
```diff
@@ -0,0 +1,85 @@
...
+}
+
+void prepare_inputs(uint8_t *input_data, uint8_t *classes)
+{
+    randombytes(input_data, number_measurements * chunk_size);
+    for (size_t i = 0; i < number_measurements; i++) {
+        classes[i] = randombit();
+        if (classes[i] == 0) {
+            *(uint16_t *) (input_data + i * chunk_size) = 0x00;
+        } else {
```
jserv: Remove non-effective `else`.

console.c
```diff
@@ -98,9 +99,11 @@ void init_cmd()
     add_cmd("log", do_log_cmd, " file           | Copy output to file");
     add_cmd("time", do_time_cmd, " cmd arg ...    | Time command execution");
     add_cmd("#", do_comment_cmd, " ...            | Display comment");
+    add_param("simulation", (int *) &simulation, "Start/Stop simulation mode",
```
jserv: The extra casting means the breakage of function prototype. Maybe we shall roll back the changes.

dudect/constant.c
```diff
@@ -0,0 +1,85 @@
...
+#include "../queue.h"
+#include "../random.h"
+#include "cpucycles.h"
+
+/* Allow random number range from 0 to 65535 */
+const size_t chunk_size = 16;
+/* Number of measurements per test */
+const size_t number_measurements = 150;
+const int drop_size = 20;
+/* Maintain a queue independent from the qtest */
```
jserv: Explain the reason why you maintain another queue.

dudect/fixture.c
```diff
@@ -0,0 +1,275 @@
...
+extern const int drop_size;
+extern const size_t chunk_size;
+extern const size_t number_measurements;
+static t_ctx *t[number_tests];
+static int64_t percentiles[number_percentiles] = {0};
+
+// threshold values for Welch's t-test
+#define t_threshold_bananas 500  // test failed, with overwhelming probability
+#define t_threshold_moderate \
+    20  // test failed. Pankaj likes 4.5 but let's be more lenient
```
jserv: Rewrite the comment since we are going to tweak threshold.

dudect/fixture.c
```diff
@@ -0,0 +1,275 @@
...
+}
+
+static int cmp(const int64_t *a, const int64_t *b)
+{
+    return (int) (*a - *b);
+}
+
+int64_t percentile(int64_t *a, double which, size_t size)
+{
+    qsort(a, size, sizeof(int64_t), (int (*)(const void *, const void *)) cmp);
```
jserv: Why did you cast with `(int (*)(const void *, const void *))`?

dudect/fixture.c
```diff
@@ -0,0 +1,275 @@
...
+    size -= 2 * drop_size;
+    size_t array_position = (size_t)((double) size * (double) which);
+    assert(array_position < size);
+    return a[array_position + 2 * drop_size];
+}
+
+// fill percentiles.
```
jserv: Use C89-style comments, i.e., `/* ... */`.

random.c
```diff
@@ -0,0 +1,46 @@
...
+void randombytes(uint8_t *x, size_t how_much)
+{
+    ssize_t i;
+    static int fd = -1;
+
+    ssize_t xlen = (ssize_t) how_much;
+    assert(xlen >= 0);
+    if (fd == -1) {
+        for (;;) {
+            fd = open("/dev/urandom", O_RDONLY);
```
jserv: Alternatively, we can use PRNG as mimalloic does: https://github.com/microsoft/mimalloc/blob/dev/src/random.c
> We use our own PRNG to keep predictable performance of random number generation
and to avoid implementations that use a lock. We only use the OS provided 
random source to initialize the initial seeds. 

We can retrieve from `/dev/urandom` only one time.

jserv: Cc. @YuShuanHsieh for consolidating the generation of random strings.

dudect/fixture.c
afcidk: I'll change the function prototype of `cmp` to `int (*)(const void*, const void*)` instead to avoid type casting.

afcidk: I'll remove the cropping step before applying statistical test since it does not seem to help us lower down the error.

PR #23, opened 2020-02-17
Modify redefinition of strdup

chses9440611:
There are two definitions of strdup in harness.h
and string2.h. This will cause a redefined error
when compile. Using macro #undef to undef the
strdup defined in string2.h:1291.

### DIFF
--- a/harness.h
+++ b/harness.h
@@ -63,6 +63,9 @@ void trigger_exception(char *msg);
 /* Tested program use our versions of malloc and free */
 #define malloc test_malloc
 #define free test_free
+
+/* Use undef to avoid strdup redefined error */
+#undef strdup
 #define strdup test_strdup
 
 #endif


jserv:
Since strdup-related work was contributed by @afcidk, I would defer to him.

afcidk:
@chses9440611, may I ask your build environment and the `string2.h` source? Since I cannot reproduce the error on my machine

chses9440611:
kernel version: 4.15.0-76-generic, Ubuntu 16.04
gcc version: 7.4.0
redefintion location is at /usr/include/x86_64-linux-gnu/bits/string2.h:1291

chses9440611:
@afcidk, I think it's my own machine fault. For there exists a "string2.h" file under my /usr/include/x86_64-linux-gnu/bits/ , whcih will lead to redefinition error. I think most other machines won't have this problem. Should I close this pull request directly?

afcidk:
The problem seems to lie in glibc version. `string2.h` is removed after [glibc-2.26](https://ftp.gnu.org/gnu/libc/glibc-2.26.tar.gz). The redefinition problem is caused by `strdup` inlines in `string2.h`. (It is removed in commit [ae65d4f](https://github.com/bminor/glibc/commit/ae65d4f3c3995279ca458c460ebf8bab1885fa03#diff-02b8234420fa95ceac1b413f747533ea), so glibc version >= 2.26 should not see this problem)

I think we don't need the `#ifdef strdup` guard before `#undef`, maybe we could just leave 
```
#undef strdup
#define strdup test_strdup
```
instead.

afcidk:
> @afcidk, I think it's my own machine fault. For there exists a "string2.h" file under my /usr/include/x86_64-linux-gnu/bits/ , whcih will lead to redefinition error. I think most other machines won't have this problem. Should I close this pull request directly?

Since we do not restrict glibc's version, I think this change is suitable.

chses9440611:
So, just modify the pull request or submit another pull request ?

afcidk:
Modify the pull request. You can force push the revised commit.

jserv:
Use `git commit --amend` and/or `git rebase -i` to rework commits.

jserv:
Thank @chses9440611 for contribution!

chses9440611:
Thank @jserv @afcidk, I got a good experience.

PR #26, opened 2020-02-20
Fix the test target on macOS

### LINKED ISSUE #25: "make test" does not pass through each trace on macOS
`make test` does not pass through each trace. Instead, it just iterate with command prompt.

owlfox:
To fix #25

### DIFF
--- a/scripts/aspell-pws
+++ b/scripts/aspell-pws
@@ -120,3 +120,4 @@ setjmp
 signal
 srand
 sscanf
+macOS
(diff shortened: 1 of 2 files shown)


jserv:
Mention "macOS" rather than "Mac".

owlfox:
aspell is quite demanding, so I used " Mac OS X" instead lol

jserv:
> aspell is quite demanding, so I used "Mac OS X" instead lol

You can update file `scripts/aspell-pws`.

jserv:
About grammar: https://www.quora.com/Why-is-it-to-make-it-work-not-to-make-it-works

owlfox:
> > aspell is quite demanding, so I used "Mac OS X" instead lol
> 
> You can update file `scripts/aspell-pws`.

Beautifully done!

PR #27, opened 2020-02-21
Support colored text while reporting

pingsutw:
`\033[91m` for failure testing (red)
`\033[0m` for resetting color (white)
`\033[92m` for successful testing (green)

### DIFF
--- a/scripts/driver.py
+++ b/scripts/driver.py
@@ -16,7 +16,7 @@ class Tracer:
     verbLevel = 0
     autograde = False
     useValgrind = False
-    colorFul = False
+    colored = False
 
     traceDict = {
         1: "trace-01-ops",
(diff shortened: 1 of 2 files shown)


jserv:
Considering the use case of continuous integration (CI), we might expect the output without color. So, please think of feasible designs to allow switching between colorfulness and monocolor.

pingsutw:
@jserv, Thanks for the reply and suggestion
Add `-c` option in scripts/driver.py, '-c' means colorful
By default is monocolor

scripts/driver.py
```diff
@@ -14,6 +16,7 @@ class Tracer:
     verbLevel = 0
     autograde = False
     useValgrind = False
+    colorFul = False
```
jserv: Simply use `colored`.

scripts/driver.py
```diff
@@ -124,11 +144,12 @@ def run(self, tid=0):
 
 
 def usage(name):
-    print("Usage: %s [-h] [-p PROG] [-t TID] [-v VLEVEL] [--valgrind]" % name)
+    print("Usage: %s [-h] [-p PROG] [-t TID] [-v VLEVEL] [--valgrind] [-c]" % name)
     print("  -h        Print this message")
     print("  -p PROG   Program to test")
     print("  -t TID    Trace ID to test")
     print("  -v VLEVEL Set verbosity level (0-3)")
+    print("  -c Print colored text in terminal")
```
jserv: Change the message to `Enable colored text`.

pingsutw: Done

scripts/driver.py
pingsutw: Done

pingsutw:
Thanks for the review

PR #29, opened 2020-02-22
Extend auto-indention for generic C/C++ source

ZhuMon:
Add .c file and .hpp file


jserv:
Improve the subject of git commit(s). I would expect something like:
> Extend auto-indention for generic C/C++ source

ZhuMon:
Done, thank for suggestion.

jserv:
Thank @ZhuMon for contributing.

PR #30, opened 2020-02-23
Find the problem of driver and fix that

eric5800602:
The problem is caused by subprocess.call().
It can't revieve self.command which has been
assigned to "valgrind"+self.qtest as parameter.
Subprocess.call() will treat that as path of program.

Therefore, I change the position of self.qtest to
fix the problem.

### DIFF
--- a/scripts/driver.py
+++ b/scripts/driver.py
@@ -88,7 +88,7 @@ class Tracer:
             return False
         fname = "%s/%s.cmd" % (self.traceDirectory, self.traceDict[tid])
         vname = "%d" % self.verbLevel
-        clist = [self.command, "-v", vname, "-f", fname]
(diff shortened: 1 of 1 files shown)


eric5800602:
On the following website, I discuss this problem in detail.
[Test-and-resolve](https://hackmd.io/IAM-skcSSsW696B3JWcMXw?view#Test-and-resolve)

jserv:
This patch prevents macOS from performing autograde.
@owlfox, please revise and think of the feasible ways to enable both environments.

owlfox:
My bad! Overlooked that it was actually broken on valgrind target.
It shall now be fixed by  #33.

jserv:
Close in favor of #33

PR #31, opened 2020-02-23
Fix potential out-of-bound access in dudect

AndybnACT:
Take a look at `get_random_string(...)` function. There is a chance of buffer over read if `number_measurements` exceeds the number of strings statically declared in `random_string`.
The problem was issued by Address Sanitizer and solved by using a fix macro NR_MEASURE to unify them.

### DIFF
--- a/dudect/constant.c
+++ b/dudect/constant.c
@@ -10,16 +10,17 @@
 #include "queue.h"
 #include "random.h"
 
+#define NR_MEASURE 150
 /* Allow random number range from 0 to 65535 */
 const size_t chunk_size = 16;
 /* Number of measurements per test */
-const size_t number_measurements = 150;
+const size_t number_measurements = NR_MEASURE;
 const int drop_size = 20;
 /* Maintain a queue independent from the qtest since
  * we do not want the test to affect the original functionality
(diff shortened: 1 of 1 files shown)


jserv:
You have to explain why `150` was set.

afcidk:
`number_measurements` is indicates the number of measurements (function call) in one time. 

The size of random string and `number_measurements` could be arbitrary numbers. I've set them to 150 since we can see the results shown in CLI more frequently, but it should not affect the experiment.

If the accumulated number of measurements is larger than `enough_measurements`, this value will affect the frequency of t-test value shown to the user. However, I've set the measurements to stop after reaching `enough_measurements` so we cannot see the difference.

afcidk:
Please also change the `number_measurements` in get_random_string and the `100` in line 47 to `NR_MEASURE`, thanks!

https://github.com/sysprog21/lab0-c/blob/893ae48a4622a3953b06969383e134b4012082aa/dudect/constant.c#L34

https://github.com/sysprog21/lab0-c/blob/893ae48a4622a3953b06969383e134b4012082aa/dudect/constant.c#L47

PR #33, opened 2020-02-23
Fix driver script error caused by macOS specific changes

owlfox:
Checked both `make test`, `make valgrind` on Linux/debian10,
`make test` on macOSX 10.15.3(no valgrind support at the moment)
Sorry, I forgot to check if the valgrind target still working on Linux after the fix.


jserv:
I suppose #30 was superseded with this pull request, right?

jserv:
I suppose #30 was superseded with this pull request, right?

owlfox:
Yes.
To make it working on macOS, the argument in `clist` shall not be empty.

PR #36, opened 2020-02-28
Show the clang-format summary of files

ldotrg:
Show all files at the end of git-commit hook message.
It could avoid scroll up to find each failing file.


scripts/pre-commit.hook
```diff
@@ -63,6 +63,9 @@ for FILE in $FILES; do
     fi
 done
 
+echo "Following files need to be clean up:"
```
jserv: Fix grammar.

PR #37, opened 2020-02-29
Fix system calling issue

hfyeh:
When passing clist to subprocess.call(),
under some circumstances (unknown) will cause
File or Directory Not Found error.

I fix this by concatenate clist into string
and pass shell=True as the second argument.

### DIFF
--- a/scripts/driver.py
+++ b/scripts/driver.py
@@ -76,7 +76,7 @@ class Tracer:
         vname = "%d" % self.verbLevel
         clist = [self.command, "-v", vname, "-f", fname]
         try:
-            retcode = subprocess.call(clist)
+            retcode = subprocess.call(" ".join(clist), shell=True)
         except Exception as e:
             print("Call of '%s' failed: %s" % (" ".join(clist), e))
             return False


jserv:
How can I reproduce the problem you mentioned?

jserv:
I would like to ask @owlfox for reviewing.

owlfox:
Hi @HFYEH ,
It looks like your [branch](https://github.com/HFYEH/lab0-c/commits/bugfix/calling_system_issue) is out of sync with [upstream](https://github.com/sysprog21/lab0-c/commits/master), so I believe the issue is a duplicate with #30. Please try [syncing](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork)  your fork with this upstream repo first. 

If that didn't help, then we might need more information on how to reproduce your error to look into it.

HFYEH:
Hi, @owlfox ,
Yes, this is exact the same issue with #30.
After pull the latest version, `make valgrind` can run normally.
Thank you!

jserv:
Close per @hfyeh reported.

PR #38, opened 2020-03-11
Fix the unknown option in Valgrind

ZhuMon:
While using massif in valgrind, like below
```
valgrind --tool=massif ./qtest
```

It would return `Unknown option: --show-leak-kinds=all`
In fact, the option `show-leak-kinds` should be called in `memcheck`
Just like `memcheck:leak-check=full`

So, change `--show-leak-kinds=all` to `--memcheck:show-leak-kinds=all`

### DIFF
--- a/.valgrindrc
+++ b/.valgrindrc
@@ -1,4 +1,4 @@
 --quiet
 --memcheck:leak-check=full
---show-leak-kinds=all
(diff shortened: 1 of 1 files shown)


jserv:
Thank @ZhuMon for contributing.

PR #39, opened 2020-03-12
Refine statement-like macro definitions

IepIweidieng:
### Remove the trailing `;` in the definition of function-like macros.

Programmers who are unaware that these "functions" are actually macros might write something seemly innocent such as:
```cpp
if (remove_front)
    dut_insert_head(s, n);
else
    dut_insert_tail(s, n);
```

But it will then expand to something disastrous:
```cpp
if (remove_front)
    do {                         \
        int j = n;               \
        while (j--)              \
            q_insert_head(q, s); \
    } while (0);;
else
    do {                         \
        int j = n;               \
        while (j--)              \
            q_insert_tail(q, s); \
    } while (0);;
```
(These backslashes at line ends are keeped for readability.
 In reality, these backslashes will be removed at translation phase 2, before the preprocessor is executed at translation phase 4)

Note that how the semicolon `;` the programmers put after `dut_insert_head()` terminates the `if` block, leaving the `else` dangling.

The idiom `do { } while (0)` was meant for dealing with this issue, but the trailing semicolon ruined it.

The removal of the trailing semicolon fixes this issue.

### Rewrite the definition of some function-like macros into expression.

This avoids the aforementioned issue and makes constructs such as `dut_new(), dut_free()` valid.

There are 2 reasons why these macros are defined like this:
```cpp
    #define dut_new() ((void) (q = q_new(q)))
```

1. To prevent invoking functions without parentheses.

If the macro were defined just as `(q = q_new(q))`, the programmers could write `q_size dut_new()` and later confuse themselves.

A cast to `void` prevents such problem, since a function cannot be invoked with any `void` arguments.

(The function `q_free()` already returns `void`, so the cast in the macro `dut_free()` is actually redundant)

2. To ensure that the cast occurs before suffix operations on the macro.

If the macro were defined as `(void) (q = q_new(q))`, the programmers could write `dut_new()->head++` which compiles.
Later when they rewrite it into `++dut_new()->head`, they will be confused, because the rewritten one does not compile.

A pair of parentheses enclosing the cast avoids such possibility by making suffix operations on the macro to occur after the cast.

### DIFF
--- a/dudect/constant.h
+++ b/dudect/constant.h
@@ -2,35 +2,29 @@
 #define DUDECT_CONSTANT_H
 
 #include <stdint.h>
-#define dut_new()    \
-    {                \
-        q = q_new(); \
-    }
+#define dut_new() ((void) (q = q_new()))
 
 #define dut_size(n)                                \
     do {                                           \
         for (int __iter = 0; __iter < n; ++__iter) \
             q_size(q);                             \
-    } while (0);
+    } while (0)
 
 #define dut_insert_head(s, n)    \
     do {                         \
         int j = n;               \
         while (j--)              \
             q_insert_head(q, s); \
-    } while (0);
+    } while (0)
 
 #define dut_insert_tail(s, n)    \
     do {                         \
         int j = n;               \
         while (j--)              \
             q_insert_tail(q, s); \
-    } while (0);
+    } while (0)
 
-#define dut_free() \
-    {              \
-        q_free(q); \
-    }
+#define dut_free() ((void) (q_free(q)))
 
 void init_dut();
 void prepare_inputs(uint8_t *input_data, uint8_t *classes);


jserv:
I defer to @afcidk, who was the original contributor for `dudect` integration.

afcidk:
Thanks for pointing out the redundant semicolon in `dut_insert_tail` and `dut_insert_head`!

I agree that we should replace curly brackets to square brackets to support statements such as `q_new()->head`.
However, I still have some problems about the revision to `dut_new` and `dut_free`.

I'm not sure if programmers should write `qsize_t dut_new()` even if they regard the macro as a function (that is, in `<type> <function>` manner without giving the variable name). 
Do you mean `queue_t *q = dut_new()`(which is `<type> <name> = <function>`) at this point? Maybe there's something I've not noticed. If that is the case, could you provide a more clearly example about this?

> If the macro were defined as (void) (q = q_new(q)), the programmers could write dut_new()->head++ which compiles.
We cannot write `dut_new()->head++` when `dut_new()` is of type void. Do you mean defining the macro as `(void*)(q = q_new(q))`, then use `((queue_t*)dut_new())->head++`?

IepIweidieng:
> [...] to support statements such as `q_new()->head`.

No. The point is to *forbid* such construct completely rather than to allow it.

The new macro definition forbids such usage.

> I'm not sure if programmers should write `q_size_t dut_new()` [...]

The example says `q_size dut_new()`, where `q_size` is a C function.

The function will be invoked if `dut_new()` were defined as `(q = q_new())`, because `q_size dut_new()` would expand into `q_size (q = q_new())`, which represents a valid function call.

The programmers should *never* write such a confusing thing.

The new macro definition does not allow such usage.

> We cannot write `dut_new()->head++` when `dut_new()` is of type void.

You can actually write `(void) (q = q_new())->head++` and pass compilation, because the suffix operators `->` and `++` have higher precedence than the cast operator `(void)`, so that the cast applies to the increased `head` rather than `dut_new()` itself.

By enclose the whole expression in parentheses, we can ensure that the cast to `void` applies to the `dut_new()` itself.

`dut_new()->head` should not be allowed since `dut_new()` is expected to behave like a function which returns `void`.

The point to rewrite `dut_new()` and `dut_free()` into expressions is to avoid the dangling else problem described above.

Since an expression only makes up one statement, a macro which expands to an expression does not causes the dangling else problem.

For example, the programmers can write this:
```cpp
if (cond)
    dut_new(), dut_free();
```
without knowing that `dut_new()` and `dut_free()` are not C functions.

afcidk:
Oh, it seems that I misunderstood your intention. Now I got your point.

It looks good to me :grinning:

jserv:
Thank @IepIweidieng for contributing to this project.

PR #40, opened 2020-05-17
Change Cppcheck version is at least 1.90

gu18168:
The current Cppcheck version check must be 1.x and at least 1.90 . However, the latest version of Cppcheck already has 2.x, so the regular expression has been modified to support versions 2.0 later.

### DIFF
--- a/scripts/pre-commit.hook
+++ b/scripts/pre-commit.hook
@@ -15,10 +15,9 @@ if [ $? -ne 0 ]; then
     echo "[!] cppcheck not installed. Unable to perform static analysis." >&2
     exit 1
 fi
-# Expected Cppcheck version is 1.90
-# FIXME: extract version number for 2.xx
-CPPCHECK_VER=$($CPPCHECK --version | sed -Ee 's/Cppcheck 1.([0-9]+)/\1/;q')
-if [ $CPPCHECK_VER -lt 90 ]; then
+# Expected Cppcheck version is at least 1.90
+CPPCHECK_VER=$($CPPCHECK --version | sed -Ee 's/Cppcheck ([0-9]).([0-9]+)/\1*100+\2/;q' | bc -l)
+if [ $CPPCHECK_VER -lt 190 ]; then
     echo "[!] cppcheck version must be at least 1.90." >&2
     echo -e "    Check 'Developer Info' for building Cppcheck from source:\n" \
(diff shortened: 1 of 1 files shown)


scripts/pre-commit.hook
```diff
@@ -15,10 +15,9 @@ if [ $? -ne 0 ]; then
     echo "[!] cppcheck not installed. Unable to perform static analysis." >&2
     exit 1
 fi
-# Expected Cppcheck version is 1.90
-# FIXME: extract version number for 2.xx
-CPPCHECK_VER=$($CPPCHECK --version | sed -Ee 's/Cppcheck 1.([0-9]+)/\1/;q')
-if [ $CPPCHECK_VER -lt 90 ]; then
+# Expected Cppcheck version is at least 1.90
+CPPCHECK_VER=$($CPPCHECK --version | sed -Ee 's/Cppcheck ([0-9]).([0-9]+)/\1*100+\2/;q' | bc -l)
```
jserv, line 19: Is it possible not to depend on `bc`?

gu18168: I have made some improvements and feel a little complicated...

scripts/pre-commit.hook
```diff
@@ -16,7 +16,8 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 # Expected Cppcheck version is at least 1.90
-CPPCHECK_VER=$($CPPCHECK --version | sed -Ee 's/Cppcheck ([0-9]).([0-9]+)/\1*100+\2/;q' | bc -l)
+N=2
+CPPCHECK_VER=$($CPPCHECK --version | sed -rEe ":r;s/\b[0-9]{1,$(($N-1))}\b/0&/;tr;s/Cppcheck ([0-9]+).([0-9]+)/\1\2/;q")
```
jserv, line 20: It looks a bit weird though.

gu18168: Yes. I used recursion to ensure that all numbers have a two-digit length so that 2.0 becomes 02.00 and the final result becomes 0200 , and then 1.90 will be 0190 .

jserv: How about splitting into two statements instead?
The latest version of Cppcheck is `2.1` though.

jserv:
The `master` branch included an ugly detection. I am about to close this pull request. Please send another pull request if you find an elegant solution. Thanks anyway.

PR #45, opened 2020-09-20
Fix bug of qtest and extend for completion

RinHizakura:
A small bug is found and fixed for the previous PR, which extend the use of linenoise to `qtest`. The bug is: a single `Ctrl+C` won't close the terminal immediately.

Also, there's some revise for the callback function `completion`  in `linenoiseSetCompletionCallback(completion)`. In the new implementation, `completion` will iterate the struct `param_ptr` and `cmd_ptr`, using them to complete input command and also parameters of command `option`. So when a new command or parameters are registered by `add_cmd` or `add_param`, it could be automatically figured out by linenoise API.

### DIFF
--- a/console.c
+++ b/console.c
@@ -552,18 +552,8 @@ int cmd_select(int nfds,
                fd_set *exceptfds,
                struct timeval *timeout)
 {
-    char *cmdline;
     int infd;
     fd_set local_readset;
-
-    while (!has_infile && !block_flag &&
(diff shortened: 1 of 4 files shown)


jserv:
I would like to ask @eecheng87 for reviewing.

eecheng87:
I have viewed your changed and it works well. Thanks for your revision.

jserv:
Thank @RinHizakura for contributing!

PR #46, opened 2020-09-27
Improve the test of trace-05

colinyoyo26:
Originally, we could pass trace-05 without setting `q->tail` correctly.


jserv:
Thank @colinyoyo26 for improving!

PR #52, opened 2020-10-13
Greatest Integer

surya0123:
program to find greatest integer number between three


jserv:
I don't think the program is related to `lab0-c` project.

surya0123:
I see
Sorry for any inconvenience caused sir
:)

PR #54, opened 2021-02-26
Fix commit hook check auto generated diffs

jw910731:
When uses `git commit -v`, there are auto generated diffs that is excluded in commit log. Thus these messages should also be excluded by the commit hook.


jserv (changes requested):
Improve the git commit message. At least, make it  grammatically correct along with the body.

jserv:
Thank @jw910731 for contributing!

PR #55, opened 2021-02-27
Fix error-prone option parsing

nickchenchj:
The behavior is undefined if the argument passed to `atoi()` is
not convertible to `int`. This issue could be solved by using
`strtol()`.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -1,5 +1,6 @@
 /* Implementation of testing code for queue code */
 
+#include <errno.h>
 #include <getopt.h>
 #include <signal.h>
 #include <spawn.h>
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Change the subject of git commit message to "Replace atoi with strtol"
That is, no backticks (`) should be used in git commit messages.

qtest.c
```diff
@@ -743,7 +745,12 @@ int main(int argc, char *argv[])
             infile_name = buf;
             break;
         case 'v':
-            level = atoi(optarg);
+            errno = 0;
+            level = strtol(optarg, &endptr, 10);
```
jserv, line 749: Reduce the scope of `endptr`. That is, rewrite as following:
```c
case 'v': {
    char *endptr;
    level = strtol(optarg, &endptr, 10);
    ...
}
```

jserv (changes requested):
The subject of this pull request was not addressing what you changed. You fixed error-prone option parsing in fact. Revise the commit message(s) accordingly.

qtest.c
```diff
@@ -742,9 +743,16 @@ int main(int argc, char *argv[])
...
             infile_name = buf;
             break;
-        case 'v':
-            level = atoi(optarg);
+        case 'v': {
+            char *endptr;
+            errno = 0;
+            level = strtol(optarg, &endptr, 10);
+            if (errno != 0 || endptr == optarg) {
+                fprintf(stderr, "FATAL: Invalid verbosity level.\n");
```
jserv, line 751: Remove "FATAL: " from the error message for consistency.

nickchenchj: Is "Improve consistency of error message" suitable for this commit?

jserv:
Use `git rebase -i` to rework the commits. Let's keep single git commit and write down better commit message.

jserv:
Thank @nickchenchj for contributing!

PR #63, opened 2022-02-16
Fix comment inconsistency

SmallHanley:
Modify q_delete_mid() comment in queue.c to make it consistent with
return type.


jserv:
Thank @SmallHanley for contributing!

PR #64, opened 2022-02-16
Add q_size

jim12312321:
add q_size function in queue.c.
use code via "https://hackmd.io/@sysprog/linux2022-lab0#%E7%89%9B%E5%88%80%E5%B0%8F%E8%A9%A6".


jserv:
Unintended bug-fixing or enhancements will not be accepted.

PR #71, opened 2022-02-18
Fix comment inconsistency

2020leon:
Modify q_delete_mid() comment in queue.h to make it consistent with return type.


jserv (changes requested):
You should mention the motivations for this change.

2020leon:
The motivation for the change is to synchronize the comments between `queue.h` and `queue.c`.

jserv:
> The motivation for the change is to synchronize the comments between `queue.h` and `queue.c`. (#63)

Improve the writing in git commit message.

PR #73, opened 2022-02-21
Fix do_delete_duplicate checking bug

kdnvt:
The q_delete_dup have the comments illustrate that it should delete
"all" node which is duplicated. However, I found that in the do_dedup
checking mechanism, it only check whether queue have duplicate string,
but didn't check whether it delete "all" node with duplicated value.
Therefore, I try to fix the mechanism, so it will check the condition.

In case some one may finish the q_delete_dup but have not finished
other function in queue.c, I avoid using function in queue.c to
implement do_dedup when I want to allocate space for a checking list.


qtest.c
```diff
@@ -493,6 +493,82 @@ static bool do_dedup(int argc, char *argv[])
         return false;
     }
 
+
```
jserv, line 496: No extra blank line required here.

qtest.c
```diff
@@ -493,6 +493,82 @@ static bool do_dedup(int argc, char *argv[])
...
+        report(
+            1,
+            "INTERNAL ERROR.  Could not allocate space for duplicate checking");
+        return false;
+    }
+    INIT_LIST_HEAD(dup_value);
+    element_t *item = NULL;
+    if (l_meta.l && !list_empty(l_meta.l)) {
+        bool last_dup = false;
+
```
jserv, line 509: Ditto.

qtest.c
```diff
@@ -493,6 +493,80 @@ static bool do_dedup(int argc, char *argv[])
...
+             * -    last_dup = match;
+             *
+             * and prevent qtest.c from commiting.
+             * I use valgrind --tool=massif to see the memory usage,
+             * the version using last_dup is better than normal version,
+             * so I consider that last_dup is needed.
+             *
+             * Finally, I decided to add a if statement with no operation
+             * to avoid cppcheck warning.
+             *
```
jserv, line 564: Improve the writing in comments. Use third person (descriptive) not first person (prescriptive).

qtest.c
```diff
@@ -493,6 +493,80 @@ static bool do_dedup(int argc, char *argv[])
...
+            bool match = !strcmp(item->value, next_item->value);
+            if (match && !last_dup) {
+                int n = strlen(item->value) + 1;
+                char *str = malloc(sizeof(*str) * n);
+                element_t *entry = malloc(sizeof(*entry));
+                if (!str || !entry) {
+                    report(1,
+                           "INTERNAL ERROR.  Could not allocate space for "
+                           "duplicate checking");
+                    if (str)
```
jserv, line 525: It is not necessary. Use plain `free(str)` instead of checking `strr`. See man page of `free` in detail.

eecheng87:
@kdnvt , please squash these four commits into one (by `git rebase -i`).

eecheng87 (commented):
I would like to confirm something. In this commit, do you want to detect if user deletes (`free`) duplicate nodes in queue.c? I didn't see the code to handle this and corresponding error messages to tell user that their `q_delete_dup` is wrong.

qtest.c
```diff
@@ -493,37 +493,136 @@ static bool do_dedup(int argc, char *argv[])
...
+            int cmp = strcmp(item->value, next_dup->value);
 
             // assume queue has been sorted
-            if (strcmp(item->value, next_item->value) == 0) {
-                report(1, "ERROR: Contain duplicate string on queue");
+            while (cmp > 0 && next_dup->list.next != dup_value) {
+                next_dup = list_first_entry(&next_dup->list, element_t, list);
+                cmp = strcmp(item->value, next_dup->value);
+            }
+            if (!cmp) {
```
kdnvt, line 603: If user doesn't delete all duplicated nodes on queue, the checking list `dup_value`, which records all of the duplicated value before `q_dedup`, will detect the remaining nodes, and send the error message `ERROR: Duplicate string remain on queue` to user.

That is, after `q_dedup`, if queue and checking list `dup_value` have the same string, it send the error message to user.

I didn't separate the cases whether the remaining nodes (with the same string on `dup_value`) on the queue is one or more.

Are these information what you'd like to confirm? or do you want to seperate two case, like in case one node remains on queue:  `[a,a,b,b,c] -> [a,b,c]`

`ERROR: No duplicate string on queue, however there is string which is once duplicated still on queue`

and multiple node's : `[a,a,b,b,c] -> [b,b,c]`

`ERROR: Contain duplicate string on queue`

eecheng87: Thanks for your explanation, I just misunderstand your commit message. 
I'll merge this after you removing meaningless comment I mentioned before. (` // set_noallocate_mode(false);`)

kdnvt: In the case  `[a,a,b,b,c] -> [a,b,c]` , `do_dedup` will return true on current version of `sysprog21/lab0-c` , which I think it should send the error message and return false.

kdnvt: Thanks for your review! I have removed the meaningless comment and squash the commits in one on last force-push. (6edd0f02d1c659ed58cd35ddd987ccb5cafd7cdd)

eecheng87 (approved):

PR #77, opened 2022-02-25
Fix implementation of function do_dedup

2020leon:
Since the original implementation only checked whether the new queue had duplicate strings but did not check if distinct strings were left, the function has been reimplemented to include both mechanisms.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -493,136 +493,87 @@ static bool do_dedup(int argc, char *argv[])
         return false;
     }
 
-    // establish checking list dup_value
-    struct list_head *dup_value = malloc(sizeof(*dup_value));
-    if (!dup_value) {
-        report(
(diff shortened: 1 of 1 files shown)


eecheng87:
@2020leon , what is the difference between this [PR](https://github.com/sysprog21/lab0-c/pull/73)?
In the initial implementation, the test case: `[a,a,b,b,c] -> [a,b,c]` will be correct.
However, in #73 , this issue was solved. I'm wondering are you try to solve this?

2020leon:
In #73, `[a,a,b,b,c] -> []` also passed the test, because it did not check if distinct strings were left. In this case, "c" must remain in the queue since it is distinct in the old queue.

eecheng87 (approved):

PR #88, opened 2022-04-02
Fix the inconsistent string comparison function

bochengC:
In the do_sort function, using the strcasecmp function to check the final list fit the requirement or not. But strcasecmp function cares about the order of a letter. So, if a list after sorted is "F d e f" error message will show. But actually it is correct order sorted by ascii code.
Replace the strcasecmp by strcmp. Strcmp(str* s1, str* s2) function compare the letter by ascii code and return >0 integer when s1's ascii code larger then s2's ascii code.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -628,7 +628,7 @@ bool do_sort(int argc, char *argv[])
             element_t *item, *next_item;
             item = list_entry(cur_l, element_t, list);
             next_item = list_entry(cur_l->next, element_t, list);
-            if (strcasecmp(item->value, next_item->value) > 0) {
+            if (strcmp(item->value, next_item->value) > 0) {
                 report(1, "ERROR: Not sorted in ascending order");
                 ok = false;
                 break;


eecheng87:
Hi @bochengC , please squash those two commits into one (by `git rebase -i`).

bochengC:
Done, please check again.

jserv (changes requested):
Don't mention the function name `do_sort` in the subject of git commit message. Instead, you should summarize what you intended to do.

bochengC:
I try to use new subject.

jserv:
@bochengC, would you rework this pull request?

bochengC:
@jserv, Yes, I did it. The 6a7750c commit is the newest version that modify the subject and squash two commits into one.

jserv:
> @jserv, Yes, I did it. The [6a7750c](https://github.com/sysprog21/lab0-c/commit/6a7750c889345a7759f492d6701e46ea2c1a2e8f) commit is the newest version that modify the subject and squash two commits into one.

At present, the subject "Fix the incorrectly showing error message" did not relieve the concerns I mentioned earlier -- it is neither informative nor meaningful to developers who wish to learn from git log. In the proposed change, string comparison function was substituted, and you should summarize in git commit message. Again, improve the subject for both pull request and git commit message.

jserv:
> @jserv , I modify the subtitle and the git messages. Thanks for your teaching.

Changing from `strcasecmp` to `strcmp` is not a matter of correctness but behavior consistency. The proposed subject was not accurate.

bochengC:
@jserv , I replace "incorrect" with "inconsistent".

jserv:
> @jserv , I replace "incorrect" with "inconsistent".

I defer to @eecheng87 for reviewing.

PR #89, opened 2022-06-19
Improve comments for list.h

steven1lung:
The keyword 'struct' is used in C instead of 'object', so using
object in comments may be misleading. The word 'structure' is also
used for comments in linux/list.h [1].

Clarify function description for list_entry(). Since we should state
what the function is doing instead of how it works. This is also how
list_entry() is commented in linux/list.h [1].

Make the style for function descriptions consistent, that some first
letter of the descriptions are capital and some are not. This commit
would make the first letter for descriptions all capital to make the
styles consistent.

[1]: https://elixir.bootlin.com/linux/v5.10/source/include/linux/list.h


Rukwezacarlos01:
Thanks for that ... But if I could ask for a big favour. I'm new to the
whole world if coding if you can and if you are willing to , please teach
me some of the tricks

PR #94, opened 2023-01-24
Avoid misused non re-entrant function

### LINKED ISSUE #91: In qtest.c, using non-reentrant function in signal handler can lead to unpredictable result.
In sigsegvhandler(), report() is called, which invocates non-reentrant stdio functions.

I check [signal-safety(7) — Linux manual page](https://man7.org/linux/man-pages/man7/signal-safety.7.html) to determine whether a function is reentrant,  and I wonder if this should be taken into consideration in this case.

In my opinion, using write() to stderr to replace stdio functions might be a solution.

eecheng87:
Close #91

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -1,5 +1,6 @@
 /* Implementation of testing code for queue code */
 
+#include <assert.h>
 #include <errno.h>
 #include <getopt.h>
 #include <signal.h>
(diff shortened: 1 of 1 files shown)


qtest.c
```diff
@@ -851,9 +852,11 @@ static void console_init()
 /* Signal handlers */
 static void sigsegvhandler(int sig)
 {
-    report(1,
-           "Segmentation fault occurred.  You dereferenced a NULL or invalid "
-           "pointer");
+    /* Avoid possible non-reentrant signal function be used in signal handler */
+    assert(write(1,
```
jserv, line 856: Why assert here?

eecheng87: To avoid unused-value warning, which might lead to compile error in this case.

jserv: Put explanations in git commit message.

PR #97, opened 2023-01-30
Change prototype of q_descend

eecheng87:
This commit changes return type of q_descend from "void" to "int". The return value is regarded as the number of elements in the queue after finishing this operation. `do_descend` needs this value to traverse the list and verify the order of the queue.

The relative answer had been updated in [lab0-c-private](https://github.com/sysprog21/lab0-c-private) and passed CI.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -747,11 +747,12 @@ static bool do_descend(int argc, char *argv[])
     error_check();
 
     if (exception_setup(true))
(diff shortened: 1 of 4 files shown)


jserv (changes requested):
Improve the grammar in both git commit messages and commits.

eecheng87:
Are they better?

PR #104, opened 2023-02-06
Support multiple queues and switching command

### LINKED ISSUE #101: Support switching between multiple queues
In conjunction with LeetCode [23. Merge k Sorted Lists](https://leetcode.com/problems/merge-k-sorted-lists/), we have to revisit the current command line interpreter. In particular, the following commands should be added or revised.
* `show` : Show the current queue ID and its contents.
* `prev` : Show the previous queue ID
* `next` : Show the next queue ID

The existing commands should perform on the current queue accordingly. After the above is completed, new command `merge` can be introduced. Consider the command sequence below:
```
new
ih RAND 10
sort
new
ih RAND 9
sort
new
ih RAND 8
sort
merge
```

Three queues are present before the command `merge`, and one queue is left as a result of its execution. Before performing `merge`, we should go through each queue to check if every queue is sorted in advance.

eecheng87:
Overview: This commit introduces "multiple queue" and provides associated switching commands: 'prev' and 'next'.

We first define a new type 'queue_chain_t', which refers to describe a doubly linked list containing all of queues. Here is the definition:

```
typedef struct {
    struct list_head chain_heads;
    int size;
} queue_chain_t;
```

'chain_heads' points to the head of a doubly linked list, which chains all of queues. 
'size' refers to the number of queues are created.

This commit also changes the definition of meta-data of each queue. Here is the modified definition:

```
typedef struct {
    struct list_head *l;
    struct list_head chain;
    int size;
    int id;
} list_head_meta_t;
```

'l' points to the head of the queue.
'chain' is used by chaining the heads of queues.
'size' refers to the length of this queue.
'id' refers to the unique identification number, which is an integer.

New command 'prev' and 'next' are self-explanatory, they help to switch current queue. This commit also allow 'show_queue' to print the ID of current queue.

Close #101

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -60,16 +60,21 @@ static int big_list_size = BIG_LIST;
 
 /* Global variables */
 
-/* List being tested */
+/* The context managing a chain of queues */
 typedef struct {
-    struct list_head *l; /**< meta data of list */
+    struct list_head *q;
+    struct list_head chain;
     int size;
-} list_head_meta_t;
+    int id;
+} qcontext_t;
 
-static list_head_meta_t l_meta;
+typedef struct {
+    struct list_head head;
+    int size;
+} queue_chain_t;
 
-/* Number of elements in queue */
-static size_t lcnt = 0;
+static queue_chain_t chain = {.size = 0};
+static qcontext_t *current = NULL;
 
 /* How many times can queue operations fail */
 static int fail_limit = BIG_LIST;
@@ -92,20 +97,38 @@ static bool do_free(int argc, char *argv[])
     }
 
     bool ok = true;
-    if (!l_meta.l)
-        report(3, "Warning: Calling free on null queue");
+    if (!chain.size || !current || !current->q) {
+        report(3,
+               "Warning: There is no available queue or calling free on null "
+               "queue");
+    }
     error_check();
 
-    if (lcnt > big_list_size)
+    if (current && current->size > big_list_size)
         set_cautious_mode(false);
-    if (exception_setup(true))
-        q_free(l_meta.l);
-    exception_cancel();
-    set_cautious_mode(true);
 
-    l_meta.size = 0;
-    l_meta.l = NULL;
-    lcnt = 0;
+    struct list_head *qnext = NULL;
+    if (chain.size > 1) {
+        qnext = ((uintptr_t) &current->chain.next == (uintptr_t) &chain.head)
+                    ? chain.head.next
+                    : current->chain.next;
+    }
+
+    if (current) {
+        list_del(&current->chain);
+
+        if (exception_setup(true))
+            q_free(current->q);
+        exception_cancel();
+        set_cautious_mode(true);
+    }
+
+    if (current) {
+        free(current);
+        chain.size--;
+        current = qnext ? list_entry(qnext, qcontext_t, chain) : NULL;
+    }
+
     show_queue(3);
 
     size_t bcnt = allocation_check();
@@ -126,18 +149,19 @@ static bool do_new(int argc, char *argv[])
     }
 
     bool ok = true;
-    if (l_meta.l) {
-        report(3, "Freeing old queue");
-        ok = do_free(argc, argv);
-    }
-    error_check();
 
     if (exception_setup(true)) {
-        l_meta.l = q_new();
-        l_meta.size = 0;
+        qcontext_t *qctx = malloc(sizeof(qcontext_t));
+
+        list_add_tail(&qctx->chain, &chain.head);
+
+        qctx->size = 0;
+        qctx->q = q_new();
+        qctx->id = chain.size++;
+
+        current = qctx;
     }
     exception_cancel();
-    lcnt = 0;
     show_queue(3);
 
     return ok && !error_check();
@@ -197,20 +221,19 @@ static bool do_ih(int argc, char *argv[])
         inserts = randstr_buf;
     }
 
-    if (!l_meta.l)
+    if (!current || !current->q)
         report(3, "Warning: Calling insert head on null queue");
     error_check();
 
-    if (exception_setup(true)) {
+    if (current && exception_setup(true)) {
         for (int r = 0; ok && r < reps; r++) {
             if (need_rand)
                 fill_rand_string(randstr_buf, sizeof(randstr_buf));
-            bool rval = q_insert_head(l_meta.l, inserts);
+            bool rval = q_insert_head(current->q, inserts);
             if (rval) {
-                lcnt++;
-                l_meta.size++;
+                current->size++;
                 char *cur_inserts =
-                    list_entry(l_meta.l->next, element_t, list)->value;
+                    list_entry(current->q->next, element_t, list)->value;
                 if (!cur_inserts) {
                     report(1, "ERROR: Failed to save copy of string in queue");
                     ok = false;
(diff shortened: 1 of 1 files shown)


eecheng87:
I'll integrate LeetCode [23. Merge k Sorted Lists](https://leetcode.com/problems/merge-k-sorted-lists/) into lab0-c after this commit is merged.

jserv:
Instead of defining `lists_meta_t`, how about `queue_chain_t`? It would mimic a blockchain's structure in that each block in the chain contains a set of messages (transactions). The messages in a block are mapped to the encapsulated elements in a queue, and a queue in this project can be equivalent to a block.

See the diagram below.
Blockchain
> Credit: NIST

jserv:
By the way, merely because a queue contains elements from a linked list does not imply that the queue itself is a linked list. Instead, a queue serves as a high-level data abstraction, and queues are constructed using the `struct list_head`. As a result, we won't use the word "list" to describe a chain of queues.

qtest.c
```diff
@@ -60,16 +60,20 @@ static int big_list_size = BIG_LIST;
...
-    struct list_head *l; /**< meta data of list */
+    struct list_head *l;
+    struct list_head chain;
     int size;
+    int id;
 } list_head_meta_t;
 
-static list_head_meta_t l_meta;
+typedef struct {
+    struct list_head chain_heads;
```
jserv, line 71: `chain_heads` is a bit confusing. It refers to the abstraction of the queue implementation.

qtest.c
```diff
@@ -92,20 +96,42 @@ static bool do_free(int argc, char *argv[])
...
-        q_free(l_meta.l);
-    exception_cancel();
-    set_cautious_mode(true);
 
-    l_meta.size = 0;
-    l_meta.l = NULL;
-    lcnt = 0;
+    struct list_head *_next;
+    if (lsm.size > 1) {
+        _next = ((unsigned long) &current_list->chain.next ==
```
jserv, line 111: If you want to compare memory addresses, cast them to `uintptr_t`. Also, you may use `memcmp` as well.

qtest.c
```diff
@@ -60,16 +60,20 @@ static int big_list_size = BIG_LIST;
...
 
-static list_head_meta_t l_meta;
+typedef struct {
+    struct list_head chain_heads;
+    int size;
+} queue_chain_t;
 
-/* Number of elements in queue */
-static size_t lcnt = 0;
+static queue_chain_t lsm = {.size = 0};
```
jserv, line 75: `lsm` sounds weird. You can simply call it `qchain` or `chain`.

qtest.c
```diff
@@ -60,16 +60,20 @@ static int big_list_size = BIG_LIST;
...
 /* Global variables */
 
-/* List being tested */
 typedef struct {
-    struct list_head *l; /**< meta data of list */
+    struct list_head *l;
+    struct list_head chain;
     int size;
+    int id;
 } list_head_meta_t;
```
jserv, line 68: Let's get rid of `list_head_meta_t` by appropriate abstractions.

eecheng87:
@jserv , I change the following naming:

1. `list_head_meta_t` => `node_meta_t` (drop the usage of "list". The information in this meta data is for each node in the queue)
2. `lsm` => `chain` (take your advice)
3. `current_list` => `current` (drop the usage of "list", and imitate `task_struct current` in linux kernel)
4. `chain_heads` => `chain_head` (it refers to the head of the chain)

qtest.c
```diff
@@ -92,20 +96,41 @@ static bool do_free(int argc, char *argv[])
...
-        q_free(l_meta.l);
-    exception_cancel();
-    set_cautious_mode(true);
 
-    l_meta.size = 0;
-    l_meta.l = NULL;
-    lcnt = 0;
+    struct list_head *_next;
+    if (chain.size > 1) {
+        _next = ((unsigned long) &current->chain.next ==
```
jserv, line 111: Cast to `uintptr_t` before comparing pointers.

qtest.c
```diff
@@ -92,20 +96,41 @@ static bool do_free(int argc, char *argv[])
...
+                 (unsigned long) &chain.chain_head)
+                    ? chain.chain_head.next
+                    : current->chain.next;
+    } else {
+        _next = NULL;
```
jserv, line 116: If you initialize `_next` to `NULL`, no `else` statement should appear.

jserv:
> 1. `list_head_meta_t` => `node_meta_t` (drop the usage of "list". The information in this meta data is for each node in the queue)

It is still confusing to name it `node_meta_t`. Instead, we can use generic `qcontext_t` or `context_t` for the sake of encapsulation.

qtest.c
```diff
@@ -60,16 +60,20 @@ static int big_list_size = BIG_LIST;
 
 /* Global variables */
 
-/* List being tested */
 typedef struct {
-    struct list_head *l; /**< meta data of list */
+    struct list_head *l;
```
jserv, line 64: Is it better to call it `q` which stands for queue?

qtest.c
```diff
@@ -60,16 +60,20 @@ static int big_list_size = BIG_LIST;
...
-static list_head_meta_t l_meta;
+typedef struct {
+    struct list_head chain_head;
+    int size;
+} queue_chain_t;
 
-/* Number of elements in queue */
-static size_t lcnt = 0;
+static queue_chain_t chain = {.size = 0};
+static qcontext_t *current = 0;
```
jserv, line 76: `current` should be initialized to `NULL` instead of `0`.

qtest.c
```diff
@@ -60,16 +60,20 @@ static int big_list_size = BIG_LIST;
...
 /* Global variables */
 
-/* List being tested */
 typedef struct {
-    struct list_head *l; /**< meta data of list */
+    struct list_head *l;
+    struct list_head chain;
     int size;
-} list_head_meta_t;
+    int id;
```
jserv, line 67: What if command `free` is invoked? Is `id` changed accordingly?

qtest.c
```diff
@@ -126,18 +149,19 @@ static bool do_new(int argc, char *argv[])
...
-    if (l_meta.l) {
-        report(3, "Freeing old queue");
-        ok = do_free(argc, argv);
-    }
-    error_check();
 
     if (exception_setup(true)) {
-        l_meta.l = q_new();
-        l_meta.size = 0;
+        qcontext_t *_lm = malloc(sizeof(qcontext_t));
```
jserv, line 154: You can simply call it `qctx` or `ctx`, which is better for understanding.

qtest.c
eecheng87: No. Similar to process ID, which will not change if the previous process is killed, `id` remains the same if the previous queue is freed.

qtest.c
```diff
@@ -60,16 +60,20 @@ static int big_list_size = BIG_LIST;
 
 /* Global variables */
 
-/* List being tested */
 typedef struct {
```
jserv, line 63: Add comments for `qcontext_t`.

qtest.c
```diff
@@ -60,16 +60,20 @@ static int big_list_size = BIG_LIST;
...
+    struct list_head *l;
+    struct list_head chain;
     int size;
-} list_head_meta_t;
+    int id;
+} qcontext_t;
 
-static list_head_meta_t l_meta;
+typedef struct {
+    struct list_head chain_head;
```
jserv, line 71: You can simply call it `head` rather than `chain_head` since the structure name is `queue_chain_t`.

qtest.c
```diff
@@ -92,20 +96,39 @@ static bool do_free(int argc, char *argv[])
...
         set_cautious_mode(false);
-    if (exception_setup(true))
-        q_free(l_meta.l);
-    exception_cancel();
-    set_cautious_mode(true);
 
-    l_meta.size = 0;
-    l_meta.l = NULL;
-    lcnt = 0;
+    struct list_head *_next = NULL;
```
jserv, line 109: Call it `qnext` instead of `_next`. The prefix `_` denotes internal global functions and variables. However, this variable is just local.

qtest.c
jserv: Mention the above incremental numbering in commits as well.

eecheng87: Done

qtest.c
```diff
@@ -60,16 +60,21 @@ static int big_list_size = BIG_LIST;
 
 /* Global variables */
 
-/* List being tested */
+/* Metadata for a queue */
```
jserv, line 63: Should be "The context managing a chain of queues"

PR #105, opened 2023-02-08
Add new command and associated test

eecheng87:
This commit support a new command 'merge', which merge queues into one sorted queue in ascending order. Besides, this commit forwards the declaration of 'qcontext_t' to queue.h since 'q_merge' needs to traverse the chain.

Relative test bench is added into trace-03-ops.cmd and the answer has been updated in `lab0-c-private`.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -59,14 +59,6 @@ extern int show_entropy;
 
 /* Global variables */
 
-/* The context managing a chain of queues */
-typedef struct {
-    struct list_head *q;
-    struct list_head chain;
-    int size;
-    int id;
-} queue_contex_t;
-
 typedef struct {
     struct list_head head;
     int size;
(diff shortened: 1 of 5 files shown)


qtest.c
```diff
@@ -789,6 +781,63 @@ static bool do_reverseK(int argc, char *argv[])
     return !error_check();
 }
 
+static bool do_merge(int argc, char *argv[])
```
jserv, line 784: What happens if command 'merge' is executed when all queues are not sorted?

qtest.c
```diff
@@ -789,6 +781,63 @@ static bool do_reverseK(int argc, char *argv[])
...
+    bool ok = true;
+    if (current && current->size) {
+        for (struct list_head *cur_l = current->q->next;
+             cur_l != current->q && --len; cur_l = cur_l->next) {
+            /* Ensure each element in ascending order */
+            element_t *item, *next_item;
+            item = list_entry(cur_l, element_t, list);
+            next_item = list_entry(cur_l->next, element_t, list);
+            if (strcmp(item->value, next_item->value) > 0) {
+                report(1, "ERROR: Not sorted in ascending order");
```
jserv, line 830: In this case, we shall prompt the usage of command 'sort'.

qtest.c
eecheng87: It will still perform `q_merge` but get the wrong answer, which then be detected immediately after the call.

jserv: Add the above statement in git commit messages.

queue.h
```diff
@@ -25,6 +25,21 @@ typedef struct {
     struct list_head list;
 } element_t;
 
+/**
+ * queue_contex_t - The context managing a chain of queues
+ * @q: pointer to the head of the queue
+ * @chain: used by chaining the heads of queues
+ * @size: the length of this queue
+ * @id: the unique identification number
+ *
```
jserv, line 34: Remove this unnecessary blank line.

qtest.c
eecheng87: There might be two scenarioes lead to error. First, queues are not sorted before merging. Second, there a flaw in the implementation of `q_merge`.

jserv: Then, consolidate the error messages as well.

jserv (changes requested):
Fix grammar mistakes in git commit message. Also, do follow the manner of [How to Write a Git Commit Message](https://cbea.ms/git-commit/).
* [Wrap the body at 72 characters](https://cbea.ms/git-commit/#wrap-72)

PR #107, opened 2023-02-13
Replace ternary operator to static array indexing

25077667:
The ternary operation has some branching overhead and is harder to extend fine-grained features in the future.
I propose to use static array indexing to select the data we need.
Furthermore, RESERVE the last element of the enum type as the size of an enum.

### DIFF
--- a/report.c
+++ b/report.c
@@ -55,12 +55,15 @@ void report_event(message_t msg, char *fmt, ...)
 {
     va_list ap;
     bool fatal = msg == MSG_FATAL;
-    /* clang-format off */
-    char *msg_name = (msg == MSG_WARN) ? "WARNING" :
-                                         (msg == MSG_ERROR) ? "ERROR" :
-                                                              "FATAL ERROR";
-    /* clang-format on */
-    int level = msg == MSG_WARN ? 2 : msg == MSG_ERROR ? 1 : 0;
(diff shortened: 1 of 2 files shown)


report.h
```diff
@@ -7,7 +7,7 @@
 /* Ways to report interesting behavior and errors */
 
 /* Things to report */
-typedef enum { MSG_WARN, MSG_ERROR, MSG_FATAL } message_t;
+typedef enum { MSG_WARN, MSG_ERROR, MSG_FATAL, MSG_END_ } message_t;
```
jserv, line 10: Replace `MSG_END_` with `N_MSG`, which stands for the number of available messages.

report.c
```diff
@@ -55,12 +55,17 @@ void report_event(message_t msg, char *fmt, ...)
 {
     va_list ap;
     bool fatal = msg == MSG_FATAL;
+    static const char *const msg_name_text[MSG_END_] = {
+        "WARNING",
+        "ERROR",
+        "FATAL ERROR",
+    };
+    char *msg_name = msg_name_text[2];
     /* clang-format off */
```
jserv, line 64: Remove adjacent `clang-format on/off` notation since it is not necessary.

25077667: @jserv Done.

report.h
25077667: Done.

PR #111, opened 2023-02-15
Suppress Cppcheck

komark06:
With Cppcheck 2.7, it will report below error:
    report.c:59:18: style: Variable   'msg_name_text' can be declared with const [constVariable]
    static char *msg_name_text[N_MSG] = {
Add inline suppression to suppress warning.


jserv:

PR #112, opened 2023-02-15
Make SSH_PRIVATE_KEY optional

yanjiew1:
The GitHub Action run on the fork of the original repository will not work
because the secret SSH_PRIVATE_KEY will not be set in the forked repository.

Make the step 'webfactory/ssh-agent@v0.7.0' optional to solve the problem.


yanjiew1:
GitHub Actions will fail on this pull request because
 1. secret variables (such as `SSH_PRIVATE_KEY`) are not passed into GitHub Actions that run on pull requests due to security reasons.
 2. `queue.c` does not contains the solution in this pull request

eecheng87 (approved):

PR #113, opened 2023-02-15
Support opening the GitHub action link on Linux

Cycatz:
Use "xdg-open" command to open the link


jserv:
Thank @Cycatz for contributing!

PR #114, opened 2023-02-16
Fix segfault while doing reverse K on NULL queue

yanjiew1:
In `do_reverseK`, we should check whether `current` is NULL before dereferencing `current`. We should not dereference it if it is NULL.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -775,7 +775,7 @@ static bool do_reverseK(int argc, char *argv[])
     }
 
     set_noallocate_mode(true);
-    if (exception_setup(true))
+    if (current && exception_setup(true))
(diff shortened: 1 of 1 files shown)


eecheng87:
Hi @yanjiew1 ,

There actually are several `do_.*` functions lack of null-pointer detection, can you also consolidate them in this commit?
Also, can you make the detection be consistent to other functions? For example:

```
static bool do_reverse(int argc, char *argv[])
{
...

    if (!current || !current->q)
        report(3, "Warning: Calling reverse on null queue");
...
}
```

Thanks.

yanjiew1:
Pull request #125 has already fixed this problem.

PR #124, opened 2023-02-21
Rephrase error message

WangHanChi:


jserv (changes requested):
Rephrase as following:
> At least one node violated the ordering rule.

PR #126, opened 2023-02-21
Fix some NULL pointer dereference in qtest when the queue is NULL

ItisCaleb:
There are some command don't have proper NULL check.
This commit is tend to fix it.


eecheng87:
@jserv , I think this PR is duplicated with #125

jserv:
Close in favor of #125 .

PR #127, opened 2023-02-22
Merge `do_i(h|t)` into a single function

yanjiew1:
We can merge `do_it` and `do_ih` into a single function `do_insert` and use the parameter `option` to choose whether to insert at head or tail.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -76,7 +76,11 @@ static int string_length = MAXSTRING;
 #define MIN_RANDSTR_LEN 5
 #define MAX_RANDSTR_LEN 10
 static const char charset[] = "abcdefghijklmnopqrstuvwxyz";
-
+/* For do_insert and do_remove */
+typedef enum {
+    POS_TAIL,
+    POS_HEAD,
+} position_t;
 /* Forward declarations */
 static bool q_show(int vlevel);
 
@@ -173,15 +177,16 @@ static void fill_rand_string(char *buf, size_t buf_size)
     buf[len] = '\0';
 }
 
-/* insert head */
-static bool do_ih(int argc, char *argv[])
+/* insertion */
+static bool do_insert(position_t pos, int argc, char *argv[])
 {
     if (simulation) {
         if (argc != 1) {
             report(1, "%s does not need arguments in simulation mode", argv[0]);
             return false;
         }
-        bool ok = is_insert_head_const();
+        bool ok =
+            pos == POS_TAIL ? is_insert_tail_const() : is_insert_head_const();
         if (!ok) {
             report(1,
                    "ERROR: Probably not constant time or wrong implementation");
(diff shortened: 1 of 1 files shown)


qtest.c
```diff
@@ -214,18 +215,22 @@ static bool do_ih(int argc, char *argv[])
     }
 
     if (!current || !current->q)
-        report(3, "Warning: Calling insert head on null queue");
+        report(3, option ? "Warning: Calling insert tail on null queue"
```
jserv, line 218: Can you consolidate the error message? That is, use format string rather than two distinct ones.

qtest.c
```diff
@@ -173,15 +173,16 @@ static void fill_rand_string(char *buf, size_t buf_size)
     buf[len] = '\0';
 }
 
-/* insert head */
-static bool do_ih(int argc, char *argv[])
+/* insertion */
+static bool do_insert(int option, int argc, char *argv[])
```
jserv, line 177: Use `enum` for operations.

qtest.c
```diff
@@ -76,7 +76,8 @@ static int string_length = MAXSTRING;
 #define MIN_RANDSTR_LEN 5
 #define MAX_RANDSTR_LEN 10
 static const char charset[] = "abcdefghijklmnopqrstuvwxyz";
-
+/* For do_insert and do_remove */
+enum position { tail = 0, head = 1 };
```
jserv, line 80: Rename the identifiers as `POS_TAIL` and `POS_HEAD`. You don't have to specify the values since C compilers always do for you implicitly.

jserv: Use additional `typedef`. That is,
```
enum {
    POS_TAIL,
    POS_HEAD,
} position_t;
```

qtest.c
```diff
@@ -173,15 +174,17 @@ static void fill_rand_string(char *buf, size_t buf_size)
     buf[len] = '\0';
 }
 
-/* insert head */
-static bool do_ih(int argc, char *argv[])
+/* insertion */
+static bool do_insert(enum position option, int argc, char *argv[])
```
jserv, line 178: Rename `option` to `pos` for the sake of readability.

qtest.c
```diff
@@ -173,15 +174,17 @@ static void fill_rand_string(char *buf, size_t buf_size)
...
 {
+    // option 0 is for insert head; option 1 is for insert tail
```
jserv, line 180: The comment is not useful.

qtest.c
yanjiew1: I just pushed a new revision. I used format string to consolidate the error message and also fixed the wrong warning message in `do_remove` using format string.

qtest.c
yanjiew1: I introduced the new type `enum position` to indicate the position `head` or `tail` to insert or remove.
Two functions, `do_insert` and `do_remove` now use the new type.

qtest.c
yanjiew1: OK. I've changed the declaration of that `enum`

qtest.c
yanjiew1: I've renamed `option` to `pos`

qtest.c
yanjiew1: I've removed the comments of `option` in both `do_remove` and `do_insert`.

qtest.c
jserv: Next time, you can simply click "Resolve conversation" once you ensure the completion of proposed changes.

yanjiew1: OK, thank you.

yanjiew1:
The latest force-pushed commit changed only the commit message.

qtest.c
```diff
@@ -173,15 +177,16 @@ static void fill_rand_string(char *buf, size_t buf_size)
     buf[len] = '\0';
 }
 
-/* insert head */
-static bool do_ih(int argc, char *argv[])
+/* insertion */
+static bool do_insert(position_t pos, int argc, char *argv[])
```
jserv, line 181: Avoid prefix `do_` which implies the association with command interpreter. Instead, you can say, `queue_insert` or `insert` since it is a private function.

qtest.c
```diff
@@ -263,85 +273,20 @@ static bool do_ih(int argc, char *argv[])
...
-        }
-    }
-    exception_cancel();
-    q_show(3);
-    return ok;
+    return do_insert(POS_TAIL, argc, argv);
 }
 
-static bool do_remove(int option, int argc, char *argv[])
+static bool do_remove(position_t pos, int argc, char *argv[])
```
jserv, line 288: Ditto. Drop the use of prefix `do_`.

yanjiew1: I changed `do_remove` to `queue_remove` as the name `remove` conflicts with a function in `stdio.h`

eecheng87 (approved):

PR #128, opened 2023-02-22
Fix warning message in do_remove

yanjiew1:
Before this change, removing both tail on empty queue displays the message "Warning: Calling remove head on empty queue," which is incorrect.

This change corrects the warning message.

### DIFF
--- a/qtest.c
+++ b/qtest.c
@@ -395,7 +395,8 @@ static bool do_remove(int option, int argc, char *argv[])
     removes[string_length + STRINGPAD] = '\0';
 
     if (!current || !current->size)
-        report(3, "Warning: Calling remove head on empty queue");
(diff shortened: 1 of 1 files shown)


qtest.c
```diff
@@ -395,7 +395,8 @@ static bool do_remove(int option, int argc, char *argv[])
     removes[string_length + STRINGPAD] = '\0';
 
     if (!current || !current->size)
-        report(3, "Warning: Calling remove head on empty queue");
+        report(3, option ? "Warning: Calling remove tail on empty queue"
```
jserv, line 398: Can you consolidate the error messages? Utilize format string.

yanjiew1:
The pull request #127 already contains the change in this pull request.
We can close this pull request after #127 merges.

PR #129, opened 2023-02-22
Fix typo in the name of function

komark06:
Replace line_hostory_load with line_history_load.


jserv:

PR #130, opened 2023-02-23
Fix warning message in do_descend

yanjiew1:
The warning message was wrong when calling on empty queue.
In addition, it should be "Calling **descend** on ..."


jserv:
Thank @yanjiew1 for contributing!

PR #131, opened 2023-02-28
Fix documentation in queue.h: q_reverseK returns nothing

ShamrockLee:
This PR fixes the documentation for q_reverseK in `queue.h`. It stated that this function returns the modified list, but it will return nothing (`void`) according to the documentation.

If applied, the developers would not be confused by that inconsistency.


eecheng87:
Hi @ShamrockLee , [checksum ](https://github.com/sysprog21/lab0-c/blob/master/scripts/checksums) of queue.h also needs the update.

ShamrockLee:
@eecheng87 The original hash is identical to the SHA1 digest of that file, so I updated it with the result from `sha1sum`.

PR #132, opened 2023-03-02
Fix error in make valgrind

WangHanChi:
Because "chmod u+x $(patched_file)" this command need root prtmission,
so I add "sudo" in this command.


jserv:
> Because "chmod u+x $(patched_file)" this command need root prtmission, so I add "sudo" in this command.

You should instead present rational explanation behind why superuser permission is required.

WangHanChi:
In my observation, I found that if I use "sudo make valgrind", it will cause the permissions in some files become root.
And there will be a error when make or make clean in next time.
In short, I misused the "sudo" and made this mistake.
Therefore, this pull request is invalid.

PR #134, opened 2023-03-05
Delete the ignore file

Risheng1128:
After input the command like `new`, the file `.cmd_history` is generated. However, when input the command `make clean`, it is still exist. Hence, it should be deleted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,4 +80,7 @@ clean:
 	rm -rf *.dSYM
 	(cd traces; rm -f *~)
 
+distclean: clean
+	rm -f .cmd_history
+
 -include $(deps)


Makefile
```diff
@@ -75,7 +75,7 @@ valgrind: valgrind_existence
 	@echo "scripts/driver.py -p $(patched_file) --valgrind -t <tid>"
 
 clean:
-	rm -f $(OBJS) $(deps) *~ qtest /tmp/qtest.*
+	rm -f $(OBJS) $(deps) *~ qtest /tmp/qtest.* .cmd_history
```
jserv, line 78: Move the deletion of `.cmd_history` in target `distclean` rather than `clean`.

PR #135, opened 2023-03-15
Reduce the size of the code

paul90317:
In this commit, I reduced the size of the code without changing its logic to enhance its readability. Below is what I changed.

Remove the casting `(uintptr_t)`.

Replace `if (current->size)` to `else`.

Remove the variable `tmp`.


jserv:
@paul90317, Thanks for contributing!

PR #139, opened 2023-03-30
Modify command description of ascend

kata1219:
Replace `strictly greater value` with `strictly less value`.


jserv (changes requested):
Read https://cbea.ms/git-commit/ carefully and refine the git commit message.

PR #146, opened 2023-11-11
Revert "Fix duplicate item detection issue"

eecheng87:
@ericlinsechs 
Revert this change.

Missing status of `is_this_dup` makes control flow go inside `else` statement, then trigger error. (check out `trace-06-ops.cmd`)
Could you please verify my observation if possible? Thanks.


ericlinsechs:
Hi @eecheng87 ,

Could you please giving more details on why the control flow will go to `else` if `in_this_dup` is removed?

Issue #153, opened 2024-02-20
Inconsistent Cppcheck Version Leads to Differing Default Enabled Checking Flags


25077667:
## Environment

- **Cppcheck Versions**: Discrepancy between Cppcheck 2.7 (Docker, Ubuntu 22.04) and Cppcheck 2.12.0 (Host, Archlinux)
- **Operating Systems**: Ubuntu 22.04 (Docker) vs. Archlinux (Host)
- **Issue Context**: Encountered during execution of a git pre-commit hook.

## Issue Description

The project encounters differing static analysis results when committing changes, triggered by the git pre-commit hook using `cppcheck`. The root cause appears to be the version difference between the Cppcheck tool in my Docker environment (version 2.7) and on my host system (version 2.12.0), leading to varying default enabled checking flags.

### Encountered Issues

- Different `cppcheck` warnings and errors are reported due to the version discrepancy.
- These include suggestions for declaring parameters as pointers to const, an error about returning a pointer to a local variable, style recommendations, and a notice on limited value flow analysis.

## Proposed Solution

To address this issue and achieve consistent static analysis checks across environments, I have implemented specific changes in the codebase and the static analysis configuration.

### Changes Made

1. **Code Adjustments**: Updated function parameters from `char *` to `const char *` where appropriate to align with `cppcheck` recommendations, enhancing code safety and maintainability.
2. **Pre-commit Hook Script Update**: Modified the `scripts/pre-commit.hook` to include new suppressions (`returnDanglingLifetime`, `constParameterCallback`, `constParameterPointer`) and added `--check-level=exhaustive` to the `cppcheck` options for a thorough analysis.

### Implementation Details

The modifications involve:
- Making parameters `const` in several functions across `report.c` and `report.h` files to prevent unintended modifications.
- Adjusting the pre-commit hook to tailor the `cppcheck` analysis, reducing false positives and focusing on relevant issues.
- See: https://github.com/25077667/lab0-c/commit/673f5a1559d97cc0ab542ee8c0291fc11c101a30

### Expected Outcome

With these changes, we expect to:
- Harmonize static analysis checks across different `cppcheck` versions, reducing inconsistencies.
- Improve code quality by adhering to best practices recommended by static analysis.
- Streamline the development process by minimizing disruptive and differing analysis outcomes between environments.

## Request for Feedback

I've proposed a solution that seems to address the observed issues effectively. However, I welcome any feedback or suggestions for improvement from the community. Specifically:
[...]

PR #154, opened 2024-02-21
Suppress Cppcheck constant check

25077667:
Refine the pre-commit hook to enforce a minimum Cppcheck Version of 1.90, alerting users with an error message and guidance on compiling Cppcheck from source if their version is outdated.

Additionally, upgrade the hook to include the
"--check-level=exhaustive" option for Cppcheck versions 2.11 and above, ensuring a comprehensive analysis.
This adjustment caters to the enhanced capabilities available from version 2.11, promoting more detailed code quality checks.

Eliminate the "--check-level=" is needed after 2.11.
#153

### DIFF
--- a/scripts/pre-commit.hook
+++ b/scripts/pre-commit.hook
@@ -16,7 +16,8 @@ CPPCHECK_suppresses="--inline-suppr harness.c \
 --suppress=nullPointer:qtest.c \
 --suppress=returnDanglingLifetime:report.c \
 --suppress=constParameterCallback:console.c \
---suppress=constParameterPointer:console.c"
+--suppress=constParameterPointer:console.c \
+--suppress=checkLevelNormal:log2_lshift16.h"
 CPPCHECK_OPTS="-I. --enable=all --error-exitcode=1 --force $CPPCHECK_suppresses $CPPCHECK_unmatched ."
 
 RETURN=0


scripts/pre-commit.hook
```diff
@@ -32,17 +32,21 @@ if [ $? -ne 0 ]; then
...
+
+# Check for version less than 1.90, we need at least 1.90
+if [ "$CPPCHECK_MAJOR_VERSION" -eq 1 ] && [ "$CPPCHECK_MINOR_VERSION" -lt 90 ]; then
+    echo "[!] cppcheck version must be at least 1.90." >&2
+    echo -e "    Check 'Developer Info' for building Cppcheck from source:\n" \
+            "          http://cppcheck.sourceforge.net/devinfo/" >&2
+    exit 1
+fi
+
+# Append --check-level=exhaustive if the version is 2.11+
```
jserv, line 47: You should explain the reason why option `--check-level=exhaustive` is appended.

scripts/pre-commit.hook
```diff
@@ -32,17 +32,22 @@ if [ $? -ne 0 ]; then
...
+# Check for version less than 1.90, we need at least 1.90
+if [ "$CPPCHECK_MAJOR_VERSION" -eq 1 ] && [ "$CPPCHECK_MINOR_VERSION" -lt 90 ]; then
+    echo "[!] cppcheck version must be at least 1.90." >&2
+    echo -e "    Check 'Developer Info' for building Cppcheck from source:\n" \
+            "          http://cppcheck.sourceforge.net/devinfo/" >&2
+    exit 1
+fi
+
+# We need to append the --check-level=exhaustive flag becuase cppcheck would complain: 
+# "log2_lshift16.h:15.1 ValueFlow analysis is limited in log2_lshift16. Use --check-level=exhaustive if full analysis is wanted."
```
jserv, line 48: It is not necessary to validate log2_lshift16.h because this file was primarily intended to illustrate the concept behind logarithm calculation. In other words, its omission was intentional. I don't expect someone would modify this file.

jserv: Alternatively, skip the checking of file `log2_lshift16.h`.

25077667: @jserv It seems hard to manually select log2_lshift16.h for out current script. It might takes a large changes commit. So, I suppress checkLevelNormal for `log2_lshift16.h`.

jserv (changes requested):
Improve git commit message to reflect what you proposed to change.

25077667:
@jserv What's about this title?

jserv (changes requested):
Make git commit message informative.

25077667:
@jserv Is this okay? "Suppress Cppcheck constant check"

jserv:
> @jserv Is this okay? "Suppress Cppcheck constant check"

The subject is acceptable, but there is no commit message body.

PR #160, opened 2024-02-25
Create c-cpp.yml

beginnerinphysics:
in order to run make, add a workflow in repository forked from jserv lab0-c


beginnerinphysics:
test

PR #162, opened 2024-02-29
Integrate linenoise with web server

vax-r:
## Summary

In the previous version of console implementation, we tried to integrate tiny-web-server to enable the ability of processing commands from web requests. As the result, the package linenoise which is responsible for
command-line auto-complete needs to be disabled during the running time of tiny-web-server. Because the `line_edit()` function in linenoise.c doesn't have the ability to handle web requests correctly.

When we start the web server, we use `cmd_select` in console.c and use `select` system call to monitor web socket file descriptor and stdin_fd at the same time, however, this ability should present in the `line_edit` function in linenoise.c so we can process commands from command-line and from web requests at the same time.

That's the reason I re-design the linenoise implementation and make some modification to put `select()` system call in `line_edit` so we can have the full ability to use web server and linenoise package in command-line at the same time.

### DIFF
--- a/linenoise.c
+++ b/linenoise.c
@@ -124,6 +124,7 @@ static char *unsupported_term[] = {"dumb", "cons25", "emacs", NULL};
 static line_completion_callback_t *completion_callback = NULL;
 static line_hints_callback_t *hints_callback = NULL;
 static line_free_hints_callback_t *free_hints_callback = NULL;
+static line_eventmux_callback_t *eventmux_callback = NULL;
 
 static struct termios orig_termios; /* In order to restore at exit.*/
 static bool maskmode = false; /* Show "***" instead of input. For passwords. */
@@ -470,6 +471,20 @@ void line_set_hints_callback(line_hints_callback_t *fn)
 void line_set_free_hints_callback(line_free_hints_callback_t *fn)
 {
     free_hints_callback = fn;
+}
+
+/* Register a function to perform I/O multiplexing to monitor multiple file
+ * descriptor from different input at the same time, so we can allow the ability
+ * of receiving commands from different input sources and still provides the
+ * command-line auto-complete feature of this package. For example, the main
+ * loop of this package can only deal with standard input file descriptor
+ * originally. When this callback function is invoked, it allows the main loop
+ * of this package to deal with multiple file descriptors from different input
+ * alongside with the origin feature to deal with the standard input.
+ */
+void line_set_eventmux_callback(line_eventmux_callback_t *fn)
+{
+    eventmux_callback = fn;
 }
 
 /* This function is used by the callback function registered by the user
@@ -931,6 +946,12 @@ static int line_edit(int stdin_fd,
         signed char c;
         int nread;
         char seq[5];
+
+        if (eventmux_callback != NULL) {
+            int result = eventmux_callback(l.buf);
+            if (result != 0)
+                return result;
+        }
 
         nread = read(l.ifd, &c, 1);
         if (nread <= 0)
--- a/linenoise.h
+++ b/linenoise.h
@@ -54,9 +54,11 @@ typedef struct {
 typedef void(line_completion_callback_t)(const char *, line_completions_t *);
 typedef char *(line_hints_callback_t)(const char *, int *color, int *bold);
 typedef void(line_free_hints_callback_t)(void *);
+typedef int(line_eventmux_callback_t)(char *);
 void line_set_completion_callback(line_completion_callback_t *);
 void line_set_hints_callback(line_hints_callback_t *);
 void line_set_free_hints_callback(line_free_hints_callback_t *);
+void line_set_eventmux_callback(line_eventmux_callback_t *);
 void line_add_completion(line_completions_t *, const char *);
 /* clang-format on */
 
--- a/web.c
+++ b/web.c
@@ -23,6 +23,8 @@
 #define TCP_CORK TCP_NOPUSH
 #endif
 
+static int server_fd;
+
 typedef struct {
     int fd;            /* descriptor for this buf */
     int count;         /* unread byte in this buf */
@@ -153,6 +155,9 @@ int web_open(int port)
     /* Make it a listening socket ready to accept connection requests */
     if (listen(listenfd, LISTENQ) < 0)
         return -1;
+
+    server_fd = listenfd;
+
     return listenfd;
 }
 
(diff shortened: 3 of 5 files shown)


jserv (changes requested):
Use `git rebase` instead of `git merge`.

linenoise.h
```diff
@@ -60,6 +60,7 @@ void line_set_free_hints_callback(line_free_hints_callback_t *);
 void line_add_completion(line_completions_t *, const char *);
 /* clang-format on */
 
+void linenoise_webfd(int fd);
```
jserv, line 63: Don't change the public header of linenoise. We would stick to the APIs defined by linenoise upstream.

vax-r: Ok, then is it ok to modify the header file of web ? I need a way to access the web file descriptor in linenoise.

jserv (changes requested):
Refine the subject of git commit messages: make it compliant with https://cbea.ms/git-commit/

linenoise.c
```diff
@@ -927,6 +931,37 @@ static int line_edit(int stdin_fd,
 
     if (write(l.ofd, prompt, l.plen) == -1)
         return -1;
+
+    fd_set set;
```
jserv, line 935: Minimize the changes against `linenoise.c`, which is considered as the work of upstream. Instead, specify hooks for the same purpose.

linenoise.c
```diff
@@ -180,6 +183,7 @@ static void line_atexit(void);
 int line_history_add(const char *line);
 static void refresh_line(struct line_state *l);
 
+
```
jserv, line 186: Don't change this line. Minimize the necessary changes.

linenoise.c
```diff
@@ -105,17 +105,20 @@
 
 #include <ctype.h>
 #include <errno.h>
+#include <netinet/in.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/ioctl.h>
+#include <sys/select.h>
```
jserv, line 114: Don't change the inclusion.

linenoise.c
```diff
@@ -1215,6 +1224,7 @@ char *linenoise(const char *prompt)
     return strdup(buf);
 }
 
+
```
jserv, line 1227: Ditto. Minimize the changes.

jserv:
Remove "Successfully" from the subject of git commit messages. Make it informative.

linenoise.c
```diff
@@ -927,6 +928,11 @@ static int line_edit(int stdin_fd,
 
     if (write(l.ofd, prompt, l.plen) == -1)
         return -1;
+
+    int result = web_select(l.buf, l.ifd);
```
jserv, line 932: The proposed change lacks refinement, as it tightly integrates web server operations with the linenoise package. Linenoise is intended to be a streamlined and functional package, offering basic command line editing capabilities universally. Therefore, it's advisable to incorporate the function hook within the main loop of linenoise. This strategy will enable a standard method for callback registration, allowing web server interactions to be executed alongside linenoise's main loop, thus achieving a more decoupled approach.

vax-r: My understanding is that the main loop of linenoise is the same as the function `linenoise()` in linenoise.c , can I say that ?

jserv: > My understanding is that the main loop of linenoise is the same as the function `linenoise()` in linenoise.c , can I say that ?

No, the main loop of linenoise package appears in `line_edit` function, and you check the callbacks such as `completion_callback` for the execution of user-defined functions. Always program in an elegant way.

linenoise.c
```diff
@@ -472,6 +473,14 @@ void line_set_free_hints_callback(line_free_hints_callback_t *fn)
     free_hints_callback = fn;
 }
 
+/* Register the web select function to monitor web socket file descriptor and
```
jserv, line 476: Don't mention web related messages. Make it generic when possible.

linenoise.c
```diff
@@ -932,6 +941,12 @@ static int line_edit(int stdin_fd,
         int nread;
         char seq[5];
 
+        if (select_callback != NULL) {
```
jserv, line 944: Would it be sufficiently clear to name it `select_callback`? It might confuse someone already familiar with linenoise's internals, particularly its "selected mode." In this case, you should consider refining the name of the newly-added callback to prevent ambiguity.

linenoise.c
```diff
@@ -472,6 +473,14 @@ void line_set_free_hints_callback(line_free_hints_callback_t *fn)
     free_hints_callback = fn;
 }
 
+/* Register the select function to monitor multiple file descriptor from
```
jserv, line 476: In the context of line editing, the operation "select" is a feature, and the changes you've made pertain to multiplexing. In other words, you may choose a different term besides "select" for clearer explanation.

jserv (changes requested):
Avoid using [backtick](https://en.wikipedia.org/wiki/Backtick) in git commit message for the sake of maximum editor compatibility. You can use (double) quotation marks instead.

linenoise.c
jserv: Further refinement is required.

vax-r: Sorry I didn't noticed this part yesterday, I've made some adjustment.
I choose the word "multiplexing" and looked for its definition in [dictionary](https://dictionary.cambridge.org/zht/%E8%A9%9E%E5%85%B8/%E8%8B%B1%E8%AA%9E/multiplexing) , I think the term fits our situation here.

linenoise.c
```diff
@@ -472,6 +473,14 @@ void line_set_free_hints_callback(line_free_hints_callback_t *fn)
     free_hints_callback = fn;
 }
 
+/* Register a function to perform I/O multiplexing to monitor multiple file
+ * descriptor from different input at the same time.
+ */
+void line_set_multiplexing_callback(line_multiplexing_callback_t *fn)
```
jserv, line 479: Replace "multiplexing" with "eventmux," which stands for the multiplexor of incoming events.

web.h
```diff
@@ -9,4 +9,6 @@ char *web_recv(int fd, struct sockaddr_in *clientaddr);
 
 void web_send(int out_fd, char *buffer);
 
-#endif
+int web_select(char *buf);
```
jserv, line 12: Refine the naming.

linenoise.c
```diff
@@ -472,6 +473,14 @@ void line_set_free_hints_callback(line_free_hints_callback_t *fn)
     free_hints_callback = fn;
 }
 
+/* Register a function to perform I/O multiplexing to monitor multiple file
```
jserv, line 476: Improve the descriptions by explaining the functionality of this callback invoked inside the main loop of linenoise. You should discuss the use case for example.

jserv (changes requested):
Refine the subject "Integrate linenoise package with web server," which is a bit misleading. In fact, web server has been integrated into the whole project, and the proposed change is not exactly referred as the integration effort. Instead, it was meant to inject the main loop of line editing by introducing an extra callback which can interoperate with the web server.

web.c
```diff
@@ -23,6 +23,8 @@
 #define TCP_CORK TCP_NOPUSH
 #endif
 
+int server_fd;
```
jserv, line 26: If this variable is not intended to be public, declare it as a static storage.

jserv:
This patch looks good to me. I defer to @eecheng87 for confirmation.

vax-r:
> This patch looks good to me. I defer to @eecheng87 for confirmation.

Can I mention this PR in my homework-1 note ? or I should refine my commit in my personal repository and mentioned it instead ?

jserv:
> Can I mention this PR in my homework-1 note ? or I should refine my commit in my personal repository and mentioned it instead ?

Go ahead. Don't go off-topic here.

jserv:
Thank @vax-r for contributing!

PR #168, opened 2024-03-06
Remove unused variable and fix command prompt display issue 

millaker:
Commit ("Remove unused variable") removes an unused variable in `console.c`. `prompt_flag` in `console.c` is specified as static and the only usage is the if statement at line 583 and the only assignment to it is at line 588 which assigns 1 to `prompt_flag`. Since no other code can modify this variable, it can be removed.

Commit ("Fix command prompt showing when using file input") fixes command prompt display issue when using file as input. Commit [b13335b](https://github.com/sysprog21/lab0-c/commit/b13335b841d9fc482f532f441d64e7729594d204) ("Introduce eventmux callback function for linenoise") removes `set_echo(0)` when `infd` is not the standard input, which will show command prompt while executing command from input files. Adding it back fixes the issue.

### DIFF
--- a/console.c
+++ b/console.c
@@ -586,6 +586,7 @@ static int cmd_select(int nfds,
                 interpret_cmd(cmdline);
             fflush(stdout);
         } else {
+            set_echo(0);
             char *cmdline = readline();
             if (cmdline)
                 interpret_cmd(cmdline);
--- a/console.c
+++ b/console.c
@@ -21,7 +21,6 @@ int show_entropy = 0;
 static cmd_element_t *cmd_list = NULL;
 static param_element_t *param_list = NULL;
 static bool block_flag = false;
-static bool prompt_flag = true;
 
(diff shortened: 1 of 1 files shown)


visitorckw:
When mentioning a commit, besides referencing the commit SHA-1 ID, including the commit's subject title will make the description clearer, such as b13335b ("Introduce eventmux callback function for linenoise").

jserv:
Refine the subject of this issue, making it informative.

visitorckw:
When describing changes, it is not advisable to mention commit IDs that have not yet been merged into the upstream, as they are highly likely to be altered, rendering the description invalid. However, when referring to a commit already merged into the upstream, it should be mentioned using its SHA-1 ID along with a one-line summary of the commit.

jserv:
Split the proposed changes into two pull requests.

PR #172, opened 2024-03-16
Introduce test_calloc and alloc

komark06:
This commit introduces a generic alloc function to handle both test_malloc and test_calloc operations based on the provided alloc_func_t. Additionally, it implements the test_calloc function to allocate memory and initialize it to zero.

### DIFF
--- a/harness.c
+++ b/harness.c
@@ -2,6 +2,7 @@
 
 #include <setjmp.h>
 #include <signal.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -57,6 +58,12 @@ static int time_limit = 1;
 static jmp_buf env;
 static volatile sig_atomic_t jmp_ready = false;
 static bool time_limited = false;
+
+/* For test_malloc and test_calloc */
+typedef enum {
+    TEST_MALLOC,
+    TEST_CALLOC,
+} alloc_t;
 
 /* Internal functions */
 
@@ -115,17 +122,23 @@ static size_t *find_footer(block_element_t *b)
     return p;
 }
 
-/* Implementation of application functions */
-
-void *test_malloc(size_t size)
+static void *alloc(alloc_t alloc_type, size_t size)
 {
     if (noallocate_mode) {
-        report_event(MSG_FATAL, "Calls to malloc disallowed");
+        char *msg_alloc_forbidden[] = {
+            "Calls to malloc are disallowed",
+            "Calls to calloc are disallowed",
+        };
+        report_event(MSG_FATAL, "%s", msg_alloc_forbidden[alloc_type]);
         return NULL;
     }
 
     if (fail_allocation()) {
-        report_event(MSG_WARN, "Malloc returning NULL");
+        char *msg_alloc_failure[] = {
+            "Malloc returning NULL",
+            "Calloc returning NULL",
+        };
+        report_event(MSG_WARN, "%s", msg_alloc_failure[alloc_type]);
         return NULL;
     }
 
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Revise the subject of git commit message and pull request.

Quote from Wikipedia:
"code refactoring is the process of restructuring existing computer code — changing the factoring — without changing its external behavior."

It is evident that the external behavior is changed since the introduction of callc harness hook.

komark06:
Thank you for your review. I've updated the subject for both the git commit message and the pull request.

harness.c
```diff
@@ -58,6 +59,12 @@ static jmp_buf env;
 static volatile sig_atomic_t jmp_ready = false;
 static bool time_limited = false;
 
+/* For test_malloc and test_calloc */
+typedef enum {
+    TEST_MALLOC = FILLCHAR,
+    TEST_CALLOC = 0,
```
jserv, line 65: Swap the order of the identifiers.

harness.c
```diff
@@ -58,6 +59,12 @@ static jmp_buf env;
...
+} alloc_func_t;
```
jserv, line 66: `alloc_func_t` is misleading since the above were meant to be identifiers which can not represent the functions.

harness.c
```diff
@@ -115,17 +122,24 @@ static size_t *find_footer(block_element_t *b)
     return p;
 }
 
-/* Implementation of application functions */
-
-void *test_malloc(size_t size)
+static void *alloc(alloc_func_t func_name, size_t size)
 {
+    char *no_alloc_msg, *fail_alloc_msg;
+    if (func_name == TEST_MALLOC) {
```
jserv, line 128: Use array to aggregate the messages.

harness.c
```diff
@@ -58,6 +59,12 @@ static jmp_buf env;
 static volatile sig_atomic_t jmp_ready = false;
 static bool time_limited = false;
 
+/* For test_malloc and test_calloc */
+typedef enum {
+    TEST_MALLOC,
+    TEST_CALLOC,
+} alloc_t;
+
```
komark06, line 67: - Remove the value associated with the enum to avoid the need to change its order.
- Rename `alloc_func_t` to `alloc_t` for clarity.

harness.c
```diff
@@ -115,17 +122,19 @@ static size_t *find_footer(block_element_t *b)
...
 
-/* Implementation of application functions */
-
-void *test_malloc(size_t size)
+static void *alloc(alloc_t alloc_type, size_t size)
 {
+    char *no_alloc_msg[] = {"Calls to malloc disallowed",
+                            "Calls to calloc disallowed"};
+    char *fail_alloc_msg[] = {"Malloc returning NULL", "Calloc returning NULL"};
+
```
komark06, line 130: Utilize an array of pointers to reference various error messages

harness.c
```diff
@@ -115,17 +122,19 @@ static size_t *find_footer(block_element_t *b)
     return p;
 }
 
-/* Implementation of application functions */
-
-void *test_malloc(size_t size)
+static void *alloc(alloc_t alloc_type, size_t size)
 {
+    char *no_alloc_msg[] = {"Calls to malloc disallowed",
+                            "Calls to calloc disallowed"};
```
jserv, line 128: Append a comma (`,`) right after "Calls to calloc disallowed" to make clang-format happy.

harness.c
jserv: Ditto. Make clang-format happy.

harness.c
```diff
@@ -115,17 +122,24 @@ static size_t *find_footer(block_element_t *b)
     return p;
 }
 
-/* Implementation of application functions */
-
-void *test_malloc(size_t size)
+static void *alloc(alloc_t alloc_type, size_t size)
 {
+    char *no_alloc_msg[] = {
+        "Calls to malloc disallowed",
```
jserv, line 128: Improve the messages, making them grammatical correct.

jserv: Rename to  `msg_alloc_forbidden`.

harness.c
```diff
@@ -115,17 +122,24 @@ static size_t *find_footer(block_element_t *b)
...
+        "Calls to calloc disallowed",
+    };
+    char *fail_alloc_msg[] = {
```
jserv, line 131: Ditto. Rename to `msg_alloc_failure`.

komark06:
Improve error messaging and precisely rename variables.

jserv:
Thank @komark06 for contributing. Be aware of the amended subject.

komark06:
Thank you for your feedback.

PR #174, opened 2024-03-18
Fix description of `q_delete_dup`

lumynou5:
The originally document says `false` is returned when `head` is `NULL`, which is inconsistent with functions like `q_delete_mid` and is confirmed to be a typo.


jserv (changes requested):
Execute `sha1sum queue.h` and update the content of file `scripts/checksums`.

jserv:

PR #178, opened 2024-03-28
Remove duplicate argument check in do_size()

aftuta85:
Redundant argument check already performed in function.


jserv:
Thank @aftuta85 for contributing!

PR #191, opened 2025-02-21
Support C23 typeof operator

jserv:
The 'typeof' operator, originally a GNU extension, is now part of the C23 standard. This change enables its use in list macros where supported.

Change-Id: I6a6e4e0caa4b72f16d957e0bd87722fac1915e3d


25077667:
@jserv What do you think about the auto type placeholder since C23 (§6.7.9 Type inference)? Is it appropriate to use it here? Thanks!

jserv:
> What do you think about the auto type placeholder since C23 (§6.7.9 Type inference)? Is it appropriate to use it here? 

It sounds good. Just do it along with effective tests.

Adarshupase (approved):

lumynou5:
The `#if` directives around the difinitions of macros `list_for_each_entry` and `list_for_each_entry_safe` breaks the static analytics of code depending on the macros.  I propose to add a `type` parameter to each of the macros, so they can be used even if we don't have `typeof`, and the interfaces are also consistent with the `list_entry` family.

jserv:
@lumynou5, which version of Cppcheck are you using? I am using Cppcheck 2.17 dev (build from source) and have not seen such errors.

lumynou5:
@jserv 2.16.0 from Arch Linux official repositories.

jserv:
> Cppcheck 2.16.0 from Arch Linux official repositories.
> I tested without the `#if`'s, it passes.

You can suggest some inlined Cppcheck suppression comments for me to apply.

lumynou5:
> inlined Cppcheck suppression comments

Because the macros are not defined, `cppcheck-suppress-macro` comment cannot be used to suppress errors where the macros are used. I think we can add an `#else` branch to define dummy macros. It results CppCheck suppressed, the parameters unchanged, and code relying on the macros failed to compile -- all as before.

For example

```c
#if __LIST_HAVE_TYPEOF
#define list_for_each_entry_safe(entry, safe, head, member)            \
    for (entry = list_entry((head)->next, typeof(*entry), member),     \
        safe = list_entry(entry->member.next, typeof(*entry), member); \
         &entry->member != (head); entry = safe,                       \
        safe = list_entry(safe->member.next, typeof(*entry), member))
#else
#define list_for_each_entry_safe(entry, safe, head, member)            \
    for (entry = safe = (void *) 1;                                    \
         sizeof(struct { int : -1; });                                 \
         ++(entry), ++(safe))
#endif
```

The purpose of the assignments and the increments are to avoid CppCheck complaining potential uninitialized variables, unused variables, and can-be-constant variables.

jserv:
> Because the macros are not defined, `cppcheck-suppress-macro` comment cannot be used to suppress errors where the macros are used. I think we can add an `#else` branch to define dummy macros. It results CppCheck suppressed, the parameters unchanged, and code relying on the macros failed to compile -- all as before.

Create a new issue along with the proposed changes.

PR #209, opened 2025-02-25
Clear where non-American English appears

Dennis40816:
This PR improves the hook that detects non-American English in commit messages. 

Previously, the hook would only indicate the presence of non-American English words without specifying their locations or the words themselves. Now, the hook clearly reports the exact line numbers and problematic words, making it easier for users to identify and correct the issues.

A test is conducted below.
```
Test commit

This commit intrduces a new featre to automatially detact misspelld
words in commit mesages. 

It help users to quicky indentify and fix issuses with non-Amercan
English.  This improvment provides more clearity and acuracy in the
warnning mesages displayed.

// test order
OneAndTwo should precede TwoAndOne.

TwoAndOne should precede OneAndTwo.
```
**Before**

image

**After**

image

### DIFF
--- a/scripts/commit-msg.hook
+++ b/scripts/commit-msg.hook
@@ -124,6 +124,41 @@ read_commit_message() {
   done < $COMMIT_MSG_FILE
 }
 
+# Get positions (line, column) for each target word in a multiline string.
+# Output format: "target: line"
+get_all_match_positions() {
+  local text="$1"
+  local targets="$2"
+  local start_line=1
+  local start_col=1
+
+  while IFS= read -r target; do
+    # search for the target string
+    local result
+    result=$(
+      awk -v t="$target" -v sl="$start_line" -v sc="$start_col" '{
+        if (NR < sl) next
+        pos = index(NR == sl ? substr($0, sc) : $0, t)
+        if (pos) {
+          print NR, (NR == sl ? pos + sc - 1 : pos)
+          exit
+        }
+      }' <<< "$text"
+    )
+
+    # skip if the target is not found
+    [ -z "$result" ] && continue
+    
+    # output and update states
+    local line col
+    read -r line col <<< "$result"
+    echo "$target: $line"
+    start_line="$line"
+    start_col=$((col + 1))
+
+  done <<< "$targets"
+}
+
 #
 # Validate the contents of the commmit msg agains the good commit guidelines.
 #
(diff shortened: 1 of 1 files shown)


jserv:
Can you simplify the lengthy script?

scripts/commit-msg.hook
```diff
@@ -367,16 +402,18 @@ done
     add_warning 1 "Commit message appears to be written in Chinese: $MISSPELLED_WORDS"
   fi
 
-  # Remove quoted text and commit hashes from $FULL_COMMIT_MSG for spell checking.
-  # Handles commit references like "commit 7d05741" (short) or full 40-char hashes.
   MSG_FOR_SPELLCHECK=$(echo "$FULL_COMMIT_MSG" | sed -E \
     -e "s/(['\"][^'\"]*['\"])//g" \
     -e "s/\bcommit[[:space:]]+[0-9a-fA-F]{7,40}\b/commit/g")
-
+
```
jserv, line 408: Don't do this. Minimizes the necessary changes.

scripts/commit-msg.hook
```diff
@@ -124,6 +124,40 @@ read_commit_message() {
...
+      print NR, (NR == sl ? pos + sc - 1 : pos)
+      exit
+    }
+  }' <<< "$text"
+}
+
+#
+# Get positions (line,column) for each target word from aspell output.
+#
+
```
jserv, line 147: Ditto.

scripts/commit-msg.hook
```diff
@@ -124,6 +124,40 @@ read_commit_message() {
...
+get_all_match_positions() {
```
jserv, line 148: Can you unify the functions `get_match_position` and `get_all_match_positions`?

Dennis40816: Sure.

Dennis40816:
Which formatter is this script using?

scripts/commit-msg.hook
```diff
@@ -124,6 +124,46 @@ read_commit_message() {
   done < $COMMIT_MSG_FILE
 }
 
+#
+# Get positions (line,column) for each target word from aspell output.
+#
+#
+# Get positions (line, column) for each target word in a multiline string.
+# Output format: "target: line"
+#
```
jserv, line 133: Trim the comments. That is,
```
# Get positions (line, column) for each target word in a multiline string.
# Output format: "target: line"
```

jserv:
> Which formatter is this script using?

I have no decided yet. You can simply follow the existing indention.

Dennis40816:
Should I squash my local commits into a single commit and then force push, or simply add a new commit?

jserv:
> Should I squash my local commits into a single commit and then force push, or simply add a new commit?

Always squash for single purpose commits.

jserv:

PR #211, opened 2025-02-26
Add dummy macros for conditionally defined macros

### LINKED ISSUE #210: Add dummy macros for conditionally defined macros
Some macros are conditionally defined, such as `list_for_each_entry_safe`, which is only defined if `typeof` is supported.  CppCheck reports `unknownMacro` when these macros are used because it checks each branch of conditional compilation (as the `--force` option presents) and they are not always defined.

A possible solution is to define dummy macros to bypass the static analytics:

```c
#if __LIST_HAVE_TYPEOF
#define list_for_each_entry_safe(entry, safe, head, member)            \
    for (entry = list_entry((head)->next, typeof(*entry), member),     \
        safe = list_entry(entry->member.next, typeof(*entry), member); \
         &entry->member != (head); entry = safe,                       \
        safe = list_entry(safe->member.next, typeof(*entry), member))
#else
#define list_for_each_entry_safe(entry, safe, head, member)            \
    for (entry = safe = (void *) 1;                                    \
         sizeof(struct { int : -1; });                                 \
         ++(entry), ++(safe))
#endif
```

The assignments and increments make CppCheck consider the variables are used and suppress potential reports about uninitialized variables, unused variables, and can-be-constant variables.  The pointers with value `1` instead of `NULL` suppress reports about dereferencing null pointers.

This way, code can pass the static analytics without changes to the macro parameters.

lumynou5:
cppcheck reports "unknownMacro" when a conditionally defined macro is used.

This patch adds dummy macros to make code pass the static analytics without other effects.  The assignments and increments make cppcheck consider the variables are used to suppress potential reports about uninitialized variables, unused variables, and can-be-constant variables.  The pointers with value 1 instead of NULL suppress reports about indirectly getting value of null pointers.

Closes #210.

Change-Id: Ia8db49c0c71ab82d5da10ce389f57ebb69c99051


list.h
```diff
@@ -411,6 +411,9 @@ static inline void list_move_tail(struct list_head *node,
     for (entry = list_entry((head)->next, typeof(*entry), member); \
          &entry->member != (head);                                 \
          entry = list_entry(entry->member.next, typeof(*entry), member))
+#else
+#define list_for_each_entry(entry, head, member) \
```
jserv, line 415: Add comments for compile-time guard.

list.h
```diff
@@ -411,6 +411,10 @@ static inline void list_move_tail(struct list_head *node,
     for (entry = list_entry((head)->next, typeof(*entry), member); \
          &entry->member != (head);                                 \
          entry = list_entry(entry->member.next, typeof(*entry), member))
+#else
+// The negative width bit-field makes a compile-time error for use of this.
```
jserv, line 415: Always use C-style comments.

jserv: Mention `BUILD_BUG_ON_ZERO` macro. See
* https://elixir.bootlin.com/linux/v6.13.4/source/include/linux/build_bug.h#L16
* https://hackmd.io/@sysprog/c-bitfield

jserv:

lumynou5:
@backink I suspect it's a bug of Cppcheck while it incorrectly recognizes `int` as a label. It looks similar to [ticket #10886](https://trac.cppcheck.net/ticket/10886), but the latter was fixed 3 years ago.

backink:
Hi @lumynou5 
I solved the problem by upgrading cppcheck to the newest version, but the required version announced in the project instruction is over 1.90. And cppcheck 2.13 is the official release version of Ubuntu 24.04, which is the newest LTS version, that is, compiling cppcheck manually from source code should be added as a prerequisite to the instruction, otherwise all the participants will encounter this issue.
In my opinion, this commit should be fixed to be compatible with the default version that Ubuntu 24.04 provides.
Thanks

lumynou5:
> this commit should be fixed to be compatible with the default version that Ubuntu 24.04 provides.

I think it's not a compatibility issue, but it's indeed better to work with 2.13.0. Other than installing from other repositories or building from source, the only option is trying not to trigger the false positive. Does giving the data member a name help Cppcheck parse it correctly without other errors?

PR #213, opened 2025-03-01
Refine trace command descriptions

Dennis40816:
During testing, an error occurred in the process of executing 'trace-03-ops.cmd'. Upon investigation, it was found that the issue was caused by an incorrect implementation of 'q_reverse'. However, this function was not explicitly listed in the trace command descriptions, which could reduce debugging efficiency. Therefore, this PR updates the descriptions in the trace command files.

This PR does not modify any test data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,8 @@ Trace files
   * They are short and simple.
   * We encourage to study them to see what tests are being performed.
   * XX is the trace number (1-17).  CAT describes the general nature of the test.
+  * All functions that need to be implemented are explicitly listed.
+  * If a colon is present in the title, all functions mentioned afterwards must be correctly implemented for the test to pass.
 * `traces/trace-eg.cmd` : A simple, documented trace file to demonstrate the operation of `qtest`
 
 ## Debugging Facilities
--- a/traces/trace-01-ops.cmd
+++ b/traces/trace-01-ops.cmd
@@ -1,4 +1,4 @@
-# Test of insert_head and remove_head
+# Test of 'q_new', 'q_insert_head', and 'q_remove_head'
 option fail 0
 option malloc 0
 new
--- a/traces/trace-05-ops.cmd
+++ b/traces/trace-05-ops.cmd
@@ -1,4 +1,4 @@
-# Test of insert_head, insert_tail, remove_head, reverse, size, swap, and sort
+# Test of 'q_new', 'q_free', 'q_insert_head', 'q_insert_tail', 'q_remove_head', 'q_reverse', 'q_size', 'q_swap', and 'q_sort'
 option fail 0
 option malloc 0
 new
--- a/traces/trace-07-string.cmd
+++ b/traces/trace-07-string.cmd
@@ -1,4 +1,4 @@
-# Test of truncated strings
+# Test of truncated strings: 'q_new', 'q_free', 'q_insert_head', 'q_insert_tail', 'q_remove_head', 'q_reverse', and 'q_sort'
 option fail 0
 option malloc 0
 new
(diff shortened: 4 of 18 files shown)


jserv:
I would like to ask @lumynou5 for confirmation.

eecheng87:
The change is OK for me, but I prefer minimal changes to the purpose.

For example, just append the missing description to the original first line.
Also, I think '# Test of operations - *' is redundant since the testing index can be inferred from the file name.

traces/trace-05-ops.cmd
```diff
@@ -1,5 +1,5 @@
-# Test of insert_head, insert_tail, remove_head, reverse, size, swap, and sort
-option fail 0
```
eecheng87, line 2: I suppose this line is not comment and should not be deleted?

Dennis40816:
> The change is OK for me, but I prefer minimal changes to the purpose.
> 
> For example, just append the missing description to the original first line. Also, I think '# Test of operations - *' is redundant since the testing index can be inferred from the file name.

There were several reasons for considering the addition of a dedicated line listing the functions required for each test:  

1. To make function requirements clearer and more consistent across all test files.  
2. To avoid cluttering the test title. Placing all functions directly after the title could make it excessively long or difficult to read, especially since the title lengths vary. In my own testing, listing all functions immediately after the title made it inconvenient to read.  

I completely agree that `# Test of operations - *` is ineffective. I will soon update the format by moving the required function names to the first line and removing the operation test numbering.  

The revised format will likely be:  
**Original topic:** `'q_new', 'q_insert_head', 'q_remove_head'`  
If the original topic already contains some functions, the missing ones will be added within the topic itself:  
**Example:** `Test of insert_head and remove_head` → `Test of 'q_new', 'q_insert_head', 'q_remove_head'`.  

If you have any suggestions, feel free to share them.

traces/trace-05-ops.cmd
Dennis40816: You're right! I didn't notice that. Thanks.

traces/trace-01-ops.cmd
```diff
@@ -1,4 +1,4 @@
-# Test of insert_head and remove_head
+# Test of 'q_new', 'q_insert_head', 'q_remove_head'
```
jserv, line 1: Add proper 'and`

Dennis40816: OK.

traces/trace-01-ops.cmd
```diff
@@ -1,4 +1,4 @@
-# Test of insert_head and remove_head
+# Test of 'q_new', 'q_insert_head' and 'q_remove_head'
```
lumynou5, line 1: I suppose it would be better to add a comma before "and"?

traces/trace-07-string.cmd
```diff
@@ -1,4 +1,4 @@
-# Test of truncated strings
+# Test of truncated strings: 'q_new', 'q_free', 'q_insert_head', 'q_insert_tail', 'q_remove_head', 'q_reverse' and 'q_sort'
```
lumynou5, line 1: I think we should note the meaning (list of invoked functions) of the colons in the readme, or people might misunderstand while it also looks like "test truncated strings on each function", which doesn't make much sense.

Dennis40816: Good point.

README.md
```diff
@@ -111,6 +111,8 @@ Trace files
   * They are short and simple.
   * We encourage to study them to see what tests are being performed.
   * XX is the trace number (1-17).  CAT describes the general nature of the test.
+  * All functions that need to be implemented are explicitly listed.
+  * If a colon  is present in the title, all functions mentioned afterwards must be correctly implemented for the test to pass.
```
lumynou5, line 115:
```suggestion
  * If a colon is present in the title, all functions mentioned afterwards must be correctly implemented for the test to pass.
```

lumynou5 (changes requested):

lumynou5 (approved):

jserv:

PR #218, opened 2025-03-02
Fix incorrect comments

yy214123:
The previous comments incorrectly described the behavior using wording from q_insert_head and q_insert_tail, suggesting that the removed string would be inserted. This commit updates the descriptions to clearly explain that the string is copied to an output buffer instead.

Change-Id: Ice8c44b485ec223bf1e16b876cf53e650f1b92df


jserv:
Thank @yy214123 for contributing!

PR #219, opened 2025-03-02
Fix incorrect comments

yy214123:
The previous comment incorrectly stated that the middle node would be returned. However, the function is a boolean function that returns true or false depending on whether the deletion was successful. This commit updates the description to clarify that the node is deleted, not returned.

Change-Id: If622f30a5df04b14602b922f7dd1fb8ceabfaefd


jserv:
Thank @yy214123 for contributing!

PR #226, opened 2025-03-04
Remove similar trailer keys from commit hook

lumynou5:
The duplicated "Close" and "Closes" keys will possibly cause inconsistency between commit messages.  The one being kept is "Closes" because the implied subject is "this patch", so it should complete the following sentence:

    This patch [closes]...

just like sentences in the body of a message.

"Bug" is also removed due to its similarity to "Closes a bug".

Change-Id: Idea7c0310249c603e550a7a4e27df490efe13d4c


jserv:

PR #230, opened 2025-03-05
Clarify q_{ascend,descend} descriptions

Kent2256:
As q_ascend and q_descend require freeing the removed node, use the proper word 'delete' in their function descriptions. Remove redundant words after replacement.


jserv (changes requested):
Update the SHA-1 checksum as well.

jserv (changes requested):
Instead of "Use," explicitly mention "Clarify" in the subject.

jserv:
Thank @Kent2256 for contributing!

PR #231, opened 2025-03-05
Add missing parameter hints

Dennis40816:
The `select` command requires one parameter, but the parameter hint was missing. Referencing the `log` command, improve the parameter hint for `select`.


jserv:

PR #232, opened 2025-03-05
Avoid unnecessary spaces created by clang-format

gg21-aping:
Current `.clang_format` creates a space between a macro's name and the parenthesis, which does not follow the kernel coding style. The change is made so the unnecessary parenthesis will not be created.

Change-Id: I97bf12cd270c6aef03eca359e29636bb5fb3d616

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -24,3 +24,4 @@ ForEachMacros:
   - hlist_for_each_entry
   - rb_list_foreach
   - rb_list_foreach_safe
+SpaceBeforeParens: ControlStatementsExceptForEachMacros


.clang-format
```diff
@@ -24,3 +24,5 @@ ForEachMacros:
   - hlist_for_each_entry
   - rb_list_foreach
   - rb_list_foreach_safe
+SpaceBeforeParens: ControlStatementsExceptForEachMacros
+
```
jserv, line 28: No trailing blank line.

jserv:
Squash the comments and then force push.

jserv:
Thank @gg21-aping for contributing!

Urbaner3:
```shell
[!] queue.c does not follow the consistent coding style.

Make sure you indent as the following:
    clang-format -i queue.c

sort: -:449: disorder: oneline
```
 Maybe Hook should also be fixed in the same time. Temporarily, I commit without this PR, because I can live with the space syntax, but I can't afford that commit is not allowed.
 @gg21-aping  Please check and try to fix git hook though.

jserv:
> Maybe Hook should also be fixed in the same time. Temporarily, I commit without this PR, because I can live with the space syntax, but I can't afford that commit is not allowed. @gg21-aping Please check and try to fix git hook though.

Create a new issue along with comprehensive descriptions.

PR #235, opened 2025-03-06
Group github action log lines

SimonLiu423:
Previously, the workflow used to test the code does not group log messages, possibly making the output too lengthy and hard to scroll on the web. Especially the logs generated by dudect/fixture.c, although it shows perfectly fine on local machines by rewriting previous lines in the terminal, github actions does not seem to support it.

The log grouping is done by adding "::group::{title}" and "::endgroup::" which are mentioned in github docs. Another argument "--group-output" is added to "scripts/driver.py" to optionally enable grouping. Added another target "test_ga" in "Makefile" and replace it with "make test" in github workflow. ("ga" is the abbreviation for "github actions")

Change-Id: I77d4455967b6d7e1d118049d91e67346f6dcdc1a
Link: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines


Makefile
```diff
@@ -60,6 +60,10 @@ test: qtest scripts/driver.py
 	$(Q)scripts/check-repo.sh
 	scripts/driver.py -c
 
+test_ga: qtest scripts/driver.py
+	$(Q)scripts/check-repo.sh
+	scripts/driver.py -c --group-output
```
jserv, line 65: Don't add new target dedicated to CI pipeline. Always run tests via `test` target in all environments. See function `check_github_actions` in `scripts/common.sh` to be aware of GitHub Actions.

SimonLiu423: I used env variable `GITHUB_ACTIONS` to detect the current environment rather than depending on `check_github_actions`. Perhaps this method could also be applied throughout the project.

Makefile
```diff
@@ -62,7 +62,11 @@ check: qtest
 
 test: qtest scripts/driver.py
 	$(Q)scripts/check-repo.sh
-	scripts/driver.py -c
+	$(Q)if [ -n "$$GITHUB_ACTIONS" ]; then \
+		scripts/driver.py -c --group-output; \
+	else \
+		scripts/driver.py -c; \
+	fi
```
jserv, line 69: Don't do this. The test driver should be aware of the environments.

PR #240, opened 2025-03-07
Fix missing compiler macros in cppcheck analysis

willy-liu:
Static analysis with cppcheck fails because predefined compiler macros are not available during analysis. This causes the environment detection in the list macro to fall back to an invalid code branch, triggering false errors such as unused labels.

Although this fallback branch is not used in supported environments, it is still parsed by cppcheck unless the proper macro definitions are provided.

This commit resolves the issue by passing the necessary compiler environment macros to cppcheck through the -D option, ensuring correct evaluation of conditional logic in macro definitions.

- Prevent static analysis false positives in list macros
- Ensure the environment is correctly simulated during checks

See https://hackmd.io/@willy-liu/linux2025-homework1#q_free for details.

Change-Id: If8b51c44f194f19f109bc4aaf91a464f850e7fdc

### DIFF
--- a/scripts/pre-commit.hook
+++ b/scripts/pre-commit.hook
@@ -68,7 +68,27 @@ cppcheck_suppressions() {
   printf "%s" "$out" | sed 's/[[:space:]]*$//'
 }
 
+# Generation of standard compliance for GCC/Clang
+detect_cc_std() {
+  local STDC_VERSION=""
+  local EXTRA_DEFINES=""
+  if command -v cc >/dev/null 2>&1; then
+    if cc --version 2>/dev/null | grep -q "clang"; then
+      STDC_VERSION=$(cc -dM -E -xc /dev/null | awk '/__STDC_VERSION__/ {print $3}')
+      EXTRA_DEFINES="-D__clang__=1"
+    elif cc --version 2>/dev/null | grep -q "Free Software Foundation"; then
+      STDC_VERSION=$(cc -dM -E -xc /dev/null | awk '/__STDC_VERSION__/ {print $3}')
+      EXTRA_DEFINES="-D__GNUC__=1"
+    fi
+  fi
+  if [ -n "$STDC_VERSION" ]; then
+    EXTRA_DEFINES+=" -D__STDC__=1 -D__STDC_VERSION__=${STDC_VERSION}"
+  fi
+  echo "$EXTRA_DEFINES"
+}
+
 CPPCHECK_OPTS="-I. --enable=all --error-exitcode=1"
+CPPCHECK_OPTS+=" $(detect_cc_std)"
 CPPCHECK_OPTS+=" --force $(cppcheck_suppressions) $(cppcheck_build_unmatched)"
 CPPCHECK_OPTS+=" --cppcheck-build-dir=.out ."
 


scripts/pre-commit.hook
```diff
@@ -68,7 +68,23 @@ cppcheck_suppressions() {
...
+  EXTRA_DEFINES="-D__GNUC__=1"
+elif command -v clang >/dev/null 2>&1; then
+  STDC_VERSION=$(clang -dM -E -xc /dev/null | awk '/__STDC_VERSION__/ {print $3}')
+  EXTRA_DEFINES="-D__clang__=1"
+else
+  STDC_VERSION=""
+fi
+if [ -n "$STDC_VERSION" ]; then
+  EXTRA_DEFINES+=" -D__STDC__=1 -D__STDC_VERSION__=${STDC_VERSION}"
+fi
```
jserv, line 84: Rewrite as a bash function and then invoke it inside the assignment of `$CPPCHECK_OPTS`.

willy-liu:
Refactored the detection of compiler type and standard version
into a Bash function.

The function `get_compiler_macros` 
determines the compiler in use, extracts the
standard version, and constructs the necessary macro definitions
for static analysis tools. The detected macros are then passed
to cppcheck to ensure proper environment simulation.

- Improve code organization and clarity
- Ensure static analysis tools receive accurate macro definitions
- Support both GCC and Clang detection

scripts/pre-commit.hook
```diff
@@ -68,23 +68,25 @@ cppcheck_suppressions() {
...
-  STDC_VERSION=$(clang -dM -E -xc /dev/null | awk '/__STDC_VERSION__/ {print $3}')
-  EXTRA_DEFINES="-D__clang__=1"
-else
-  STDC_VERSION=""
-fi
-if [ -n "$STDC_VERSION" ]; then
-  EXTRA_DEFINES+=" -D__STDC__=1 -D__STDC_VERSION__=${STDC_VERSION}"
-fi
+# Check compiler type and __STDC_VERSION__
+get_compiler_macros() {
```
jserv, line 72: Rename to `generate_cc_std`, which means the generation of standard compliance for GCC/Clang.

scripts/pre-commit.hook
```diff
@@ -68,7 +68,25 @@ cppcheck_suppressions() {
...
+        EXTRA_DEFINES="-D__GNUC__=1"
+    elif command -v clang >/dev/null 2>&1; then
+        STDC_VERSION=$(clang -dM -E -xc /dev/null | awk '/__STDC_VERSION__/ {print $3}')
+        EXTRA_DEFINES="-D__clang__=1"
+    fi
+    if [ -n "$STDC_VERSION" ]; then
+        EXTRA_DEFINES+=" -D__STDC__=1 -D__STDC_VERSION__=${STDC_VERSION}"
+    fi
+    echo "$EXTRA_DEFINES"
+}
```
jserv, line 86: Use 2 spaces for indention.

willy-liu:
Standardize naming and indentation

Renamed function for consistency and adjusted indentation to align with
project style.

willy-liu:
Squashed the commits.

scripts/pre-commit.hook
```diff
@@ -68,7 +68,25 @@ cppcheck_suppressions() {
...
+detect_cc_std() {
+  local STDC_VERSION=""
+  local EXTRA_DEFINES=""
+  if command -v gcc >/dev/null 2>&1; then
+    STDC_VERSION=$(gcc -dM -E -xc /dev/null | awk '/__STDC_VERSION__/ {print $3}')
+    EXTRA_DEFINES="-D__GNUC__=1"
+  elif command -v clang >/dev/null 2>&1; then
+    STDC_VERSION=$(clang -dM -E -xc /dev/null | awk '/__STDC_VERSION__/ {print $3}')
+    EXTRA_DEFINES="-D__clang__=1"
+  fi
```
jserv, line 81: During the `make` process, compilation utilizes the `cc` command, which references either gcc or clang based on your system's configuration. On GNU/Linux systems, `cc` typically defaults to gcc, while BSD systems (including macOS) link `cc` to clang. We shall analyze the `cc` command before directly invoking either `gcc` or `clang`.

See also: https://github.com/sysprog21/rv32emu/blob/master/mk/toolchain.mk

willy-liu: After learning the `cc` command, I understand it’s a useful way to identify the C compiler. 

However, when I checked the Makefile, I noticed that `make` defaults to `gcc` because `CC = gcc` is set in the first line.

https://github.com/sysprog21/lab0-c/blob/master/Makefile

jserv: > I noticed that `make` defaults to `gcc` because `CC = gcc` is set in the first line.

The build compiler and configurations can be checked by examining the `cc` variable, as the `CC` variable may be subject to override. It's generally acceptable to do this verification.

willy-liu:
Refactored the function using `cc` to check compiler type.

jserv:
Thank @willy-liu for contributing!

PR #242, opened 2025-03-07
Introduce EditorConfig

lumynou5:
Originally, the indentation of the scripts was not unified: ones in 'scripts/' used 2 spaces, and ones in '.ci/' used 4. Even though the former used 2 spaces, the style was not implemented well as some indents in the files were 4 spaces.

This patch unifies the style of the scripts, and note it with an EditorConfig file. EditorConfig is used because it's supported out-of-the-box by many editors, helping the auto-indent functionality works.

Change-Id: I41be8779ef5b0e94192c144bb48605625727963f

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[Makefile]
+indent_style = tab
+indent_size = 4
+max_line_length = 80
+
+[**/*.[ch]]
+indent_style = space
+indent_size = 4
+max_line_length = 80
+
+[{**/*.py,.ci/*}]
+indent_style = space
+indent_size = 4
+
+[{scripts/*,.github/workflows/*}]
+indent_style = space
+indent_size = 2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,25 +10,25 @@ jobs:
     - uses: webfactory/ssh-agent@v0.7.0
       continue-on-error: true
       with:
-          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+        ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
     - name: install-dependencies
       run: |
-            .ci/check-sanity.sh
-            sudo apt-get update
-            sudo apt-get -q -y install build-essential cppcheck
+        .ci/check-sanity.sh
+        sudo apt-get update
+        sudo apt-get -q -y install build-essential cppcheck
     - name: make
       run:  |
-            git clone git@github.com:sysprog21/lab0-c-private || echo "No provisioning profile found"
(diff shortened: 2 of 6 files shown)


lumynou5:
Is it needed to separate re-formatting files and introducing EditorConfig into two commits?

.ci/check-format.sh
```diff
@@ -6,7 +6,7 @@ set -x
 
 for file in ${SOURCES};
 do
-    clang-format-18 ${file} > expected-format
-    diff -u -p --label="${file}" --label="expected coding style" ${file} expected-format
+  clang-format-18 ${file} > expected-format
```
jserv, line 9: Don't do this with this pull request.

.ci/check-newline.sh
```diff
@@ -6,14 +6,14 @@ ret=0
...
 # Reference: https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e
 while IFS= read -rd '' f; do
-    if file --mime-encoding "$f" | grep -qv binary; then
-        tail -c1 < "$f" | read -r _ || show=1
-        if [ $show -eq 1 ]; then
-            echo "Warning: No newline at end of file $f"
-            ret=1
-            show=0
-        fi
+  if file --mime-encoding "$f" | grep -qv binary; then
```
jserv, line 9: Ditto.

jserv (changes requested):
Mention [EditorConfig](https://editorconfig.org/) in top-level documentation (i.e., `README.md`).

jserv:
> Is it needed to separate re-formatting files and introducing EditorConfig into two commits?

Always minimize the necessary changes.

jserv (changes requested):
Rebase the latest `master` branch.

jserv:
Is it possible to enforce the style checks in GitHub Actions as well?

lumynou5:
> Is it possible to enforce the style checks in GitHub Actions as well?

There are some tools such as [editorconfig-check](https://github.com/editorconfig-checker/editorconfig-checker), but I think the simplest way is to run a headless editor and diff the content before and after re-indenting, for example:

```shell
cp $file $file.bak
nvim --headless +'norm vG=' +'x' $file
diff $file $file.bak
```

jserv:
> There are some tools such as [editorconfig-check](https://github.com/editorconfig-checker/editorconfig-checker), but I think the simplest way is to run a headless editor and diff the content before and after re-indenting.

It sounds great. Create an issue to enforce consistent style for the shell scripts after the merging of this pull request.

jserv:

lumynou5:
Also, EditorConfig is not language-aware. It doesn't know that Bash document-here must not be indented with spaces (or not be indented at all). Therefore, rather than enforcing checks, I recommend to let it just help editors find correct indentation and avoid errors like before this PR, and human developers can align or do whatever making it more readable to humans.

PR #244, opened 2025-03-08
Ensure repetitions input is at least 1

EricccTaiwan:
Passing 0 or a negative value for repetitions does not cause errors, as the input is an integer. However, this behavior may be confusing for users. To improve clarity, enforce repetitions to be an integer greater than or equal to 1 and return an error if the input is invalid.

Change-Id: Ifbe1e1846f6eac2c933a00a3a9af5d4c533f3c5f


jserv:

PR #247, opened 2025-03-09
Add graph plotting for dudect analysis

SimonLiu423:
Implement graph generation using "gnuplot" for dudect to visualize t-statistic measurements.

- Add "plot.c" and "plot.h" files to support graphing, "plot.c" stores each measurements execution time in "data_buffer". When plot_graph() is called, stores the data to file and generates a "gnuplot" script. I try to replicate the graph shown in "dudect's" paper "Dude, is my code constant time?".

- Modify "fixture.c" to integrate with new plotting functionality.

- Move definition of ENOUGH_MEASURE and TEST_TRIES from "fixture.c" to "fixture.h" for "plot.c" to access and specify size of "data_buffer"

- Update "Makefile" to include plot.o in build

This enhancement allows visual representation of constant-time analysis results, making it easier to interpret dudect test outcomes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,8 @@ GPATH
 GRTAGS
 GSYMS
 GTAGS
+
+# gnuplot files
+gnuplot_script.gnu
+data.txt
+*.png
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ $(GIT_HOOKS):
 	@echo
 
 OBJS := qtest.o report.o console.o harness.o queue.o \
-        random.o dudect/constant.o dudect/fixture.o dudect/ttest.o \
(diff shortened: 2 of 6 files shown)


SimonLiu423:
TODO:
- detect if gnuplot is installed on machine
- update github workflow to add support for plotting
- combine multiple plots into one single image

Known Issue:
- While generating the gnuplot script, `sprintf` is used to dynamically set the label. It is a [command](http://gnuplot.info/docs_5.5/Commands.html) provided by gnuplot. However, the `pre-commit.hook` would stop us from committing since `sprintf` is detected to be a dangerous function.

jserv:
> the pre-commit.hook would stop us from committing since sprintf is detected to be a dangerous function.

You can send another pull request which limits the detection of `sprintf` only to C source files.

PR #248, opened 2025-03-09
Fix console hang after exceeding input limit

EricccTaiwan:
When the console input exceeds the error limit (five attempts), it displays: "Error limit exceeded. Stopping command execution." However, it does not exit the command prompt. Additionally, attempting to quit using the "quit" command fails, leaving Ctrl+C as the only way to terminate the program, which may lead to memory leaks.

This commit fixes the issue by declaring do_quit before record_error and calling it upon reaching the error limit to prevent memory leaks.

Alternative solutions considered:
1. Rearranging function definitions to place do_quit before record_error, which eliminates the need for declaration but requires restructuring, a change that warrants further discussion.
2. Setting quit_flag after exceeding the limit and enforcing "quit" as the only allowed input, preserving the existing function order.

This approach balances maintainability and immediate resolution, ensuring the console exits properly once the input limit is exceeded.

Change-Id: I52d53f69156bff2ab793afba65b46c67c125671f

### DIFF
--- a/console.c
+++ b/console.c
@@ -160,12 +160,43 @@ static char **parse_args(char *line, int *argcp)
     return argv;
 }
 
+/* Handles forced console termination for record_error and do_quit */
+static bool force_quit(int argc, char *argv[])
+{
+    cmd_element_t *c = cmd_list;
+    bool ok = true;
+    while (c) {
+        cmd_element_t *ele = c;
+        c = c->next;
+        free_block(ele, sizeof(cmd_element_t));
+    }
+
+    param_element_t *p = param_list;
+    while (p) {
+        param_element_t *ele = p;
+        p = p->next;
+        free_block(ele, sizeof(param_element_t));
+    }
+
+    while (buf_stack)
+        pop_file();
+
+    for (int i = 0; i < quit_helper_cnt; i++) {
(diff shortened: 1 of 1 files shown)


console.c
```diff
@@ -160,12 +160,15 @@ static char **parse_args(char *line, int *argcp)
...
 static void record_error()
 {
     err_cnt++;
     if (err_cnt >= err_limit) {
-        report(1, "Error limit exceeded.  Stopping command execution");
-        quit_flag = true;
+        report(
+            1,
+            "Error limit exceeded.  Stopping command execution, and quitting");
+        do_quit(0, NULL);
```
jserv, line 171: Don't do this. The functions starting with `do_` are only used by the command interpreter during command dispatching. You can introduce another helper function for both `do_quit` and `record_error`.

EricccTaiwan: have introduced a helper function, `force_quit`, in the latest commit to handle both `do_quit` and `record_error`. This ensures that functions starting with `do_` remain strictly for command dispatching while reducing code duplication and improving maintainability.

jserv (changes requested):
Squash commits.

jserv:

PR #256, opened 2025-03-11
Ensure proper footer sorting of footer

charliechiou:
When no trailer was configured in git configuration, `TRAILERS_BY_REGEX` remained empty. This caused the subsequent match condition to always evaluate to true, leading to incorrect footer sorting regardless of their content. 

To address this issue, the update ensures that when the configuration is empty, default terms are used to initialize `TRAILERS_BY_REGEX`. The sorting mechanism is then adjusted to prioritize other footers, configured trailers, and the change-id.

Change-Id: I6df5438b7ec6c85f5438f869b44cd14589c03339

### DIFF
--- a/scripts/commit-msg.hook
+++ b/scripts/commit-msg.hook
@@ -130,6 +130,16 @@ get_all_match_positions() {
   done <<< "$targets"
 }
 
+KNOWN_TRAILERS_BY=(
+  'Signed-off-by'
+  'Reviewed-by'
+  'Co-authored-by'
+  'Acked-by'
+  'Suggested-by'
+  'Tested-by'
+  'Reported-by'
+)
+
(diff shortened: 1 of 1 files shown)


scripts/commit-msg.hook
```diff
@@ -151,6 +151,16 @@ build_commit_trailer_regex() {
     "Git-Dch[$separators] (Ignore|Short|Full)"
   )
 
+  default_trailers_by=(
```
jserv, line 154: Non-local variables should be capitalized. Refine `build_commit_trailer_regex` by reusing this array.

jserv (changes requested):
Always squash commits for a single purpose pull request.

jserv:
Thank @charliechiou for contributing!

PR #259, opened 2025-03-12
Fix misleading comment

Jordymalone:
The previous comments incorrectly suggested that the file descriptor was not yet ready. In fact, the check (web_fd != -1) ensures that web_fd is valid and should be monitored in the read set. This commit updates the comment to clearly state that web_fd is available, providing an accurate description of the code's behavior.

Change-Id: Ib81d417a848e1cf41f289bd2abafb2de1c8a23d8


jserv:
Thank @Jordymalone for contributing!

PR #260, opened 2025-03-12
Improve aspell dictionary

lumynou5:
This patch improves the personal dictionary used by aspell to check commit messages by
- Adding words,
- Sorting them, and
- Removing duplicated entries.

Discussion is welcome. If you find some words cannot pass the check, please leave a comment. They will be squashed into a single commit containing all added words, with "Co-authored-by" trailers for the contributors of the words.

Change-Id: I332fafc989700ea9d3dbc2fcddf746af0fa46a8d


jserv:
We should automate the sorting of words in the user-defined dictionary.

lumynou5:
> We should automate the sorting of words in the user-defined dictionary.

What do you think? Not to automate the sorting but check the order.

jserv:
> Not to automate the sorting but check the order.

It is pretty well. Show the proposed changes.

scripts/pre-commit.hook
```diff
@@ -222,6 +222,11 @@ if [ "${#SHELL_FILES[@]}" -gt 0 ]; then
   done
 fi
 
+ASPELL_DICT_FILE='scripts/aspell-pws'
+if ! tail -n +2 $ASPELL_DICT_FILE | sort -cdu; then
+  throw 'Aspell dictionary is unsorted or contains duplicated entries.'
```
jserv, line 227: Add prompt to run `sort | uniq`.

jserv:
For sorting purposes, it is generally better to ensure all keys in a user-defined dictionary are lowercase in advance. This prevents potential sorting inconsistencies since uppercase and lowercase letters have different ASCII/Unicode values, which affects their sort order. Pre-converting all keys to lowercase creates a consistent comparison basis and makes the sorting behavior more predictable and intuitive to users.

lumynou5:
> This prevents potential sorting inconsistencies since uppercase and lowercase letters have different ASCII/Unicode values, which affects their sort order.

The `-d`/`--dictionary-order` doesn't sort the entries by character codepoints and ignore the case. You can check the file:

```
aarch
abbrev
abcdefghijklmnopqrstuvwxyz
acct
AddressSanitizer
adjtime
```

Lowercase letters doesn't make entries follow ones with uppercase letters (as ASCII); instead, "Ad" is immediately following `ac`.

jserv (changes requested):
Rebase the latest `master` for the sake of force-push.

jserv:

PR #269, opened 2025-03-16
Fix missing Change-Id

### LINKED ISSUE #267: Missing Change-Id
When committing #266 , I noticed that the **Change-Id** was missing. Previously, I couldn't push such a commit to GitHub, but now I can. I'm wondering if there has been a change in the requirements or if this is a bug.

charliechiou:
The previous pull request 43a311d  mistakenly set the `trailers_by` variable to `DEFAULT_TRAILERS_BY`, causing the change ID to disappear. Change the variable to `KNOWN_TRAILERS_BY` to ensure the change ID appears correctly.
Close #267
Change-Id: I4e3b4bfa06aba31bed2c26925ecfa5593a3d327b

### DIFF
--- a/scripts/commit-msg.hook
+++ b/scripts/commit-msg.hook
@@ -186,7 +186,7 @@ build_commit_trailer_regex() {
   done < <(git config --get-regexp 'trailer.*.key')
 
   if [[ ${#trailers_by[@]} -eq 0 ]]; then
-    trailers_by=("${DEFAULT_TRAILERS_BY[@]}")
+    trailers_by=("${KNOWN_TRAILERS_BY[@]}")
   fi
 
   # Possible trailers :


jserv (changes requested):
Mention short commit hash instead of pull request numbers in commit messages. Append `Close #267` when you confirm to fix.

charliechiou:
Sorry, I'm not sure where I should append Close #267. It seems like I don’t have permission to close the issue. Thanks!

jserv:
> I'm not sure where I should append Close #267. It seems like I don’t have permission to close the issue. Thanks!

No, check this: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

jserv (changes requested):
Use your name in (Hànyǔ) Pinyin in commit messages.

jserv:
Thank @charliechiou for contributing! Do [reply the posts](https://www.facebook.com/groups/system.software2025) accordingly.

PR #271, opened 2025-03-17
Simplify progress bar display logic

EricccTaiwan:
Refactor the progress bar logic to remove redundant conditional handling for the final 100% case. Instead of using a special case to clear and reprint the final bar, the update now ensures a consistent format for all progress levels.

Change-Id: Ie424815310c18b0b021d53bf9652d590da1bf5ab

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -54,15 +54,12 @@ progress() {
   local bar_left
   bar_left=$(printf "%${left}s")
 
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Rebase the latest `master` branch, which was forcefully pushed.

jserv:

PR #274, opened 2025-03-19
Minor optimization on branch condition

Urbaner3:
Ensure that the code checks if the next character is not EOF before
attempting to check if it is a hexadecimal digit. This prevents
potential segmentation faults during list operations.

The fmtscan.c file shows error [report](https://www.facebook.com/groups/system.software2025/permalink/945106757826126/). Try to check EOF first at fmtscan.c, line 664, as 
[Check for EOF before checking isxdigit](https://github.com/sysprog21/lab0-c/commit/18218e1f7d37441019c1613d4e0da947c4d80db5) commit tells.

It turned out went well. Although at line 657 in fmtscan.c, it has checked already. I say it's the sensitive of the static analysis, the code is right in logic though. Maybe it can be optimized in condition branches. 

Cherry pick one and only related one commit, which is Check for EOF, and leave the rest as reference.


jserv:
Read [CONTRIBUTING.md](https://github.com/sysprog21/lab0-c/blob/master/CONTRIBUTING.md) carefully:
> Group Related Changes Together Each commit should encapsulate a single, coherent change. e.g., if you are addressing two separate bugs, create two distinct commits. This approach produces focused, small commits that simplify understanding, enable quick rollbacks, and foster efficient peer reviews. By taking advantage of Git’s staging area and selective file staging, you can craft granular commits that make collaboration smoother and more transparent.

Rework and submit pull requests later.

PR #280, opened 2025-03-20
Improve dudect test with cropped time analysis

BennyWang1007:
The original dudect collect all execution times and perform t-tests,
which may be affected by outliers. The outliers could be caused by
context switches, interrupts, or other system activities. This patch
introduces percentile-based cropping to remove outliers.

The patch adds a new function "prepare_percentiles()" to compute
thresholds using an complementary exponential decay scale. The function
is called before the test starts.

The patch modifies "update_statistics()" to perform t-tests on cropped
execution times by filtering out the outliers.

Change-Id: Ieba005a990b122785ce931fe1364adf81dd9802a

### DIFF
--- a/dudect/fixture.c
+++ b/dudect/fixture.c
@@ -42,6 +42,7 @@
 
 #define ENOUGH_MEASURE 10000
 #define TEST_TRIES 10
+#define NUM_PERCENTILES (100)
 
 static t_context_t *t;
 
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Check https://aeb.win.tue.nl/linux/lk/lk-8.html carefully:
> [Use the body to explain what and why vs. how](https://cbea.ms/git-commit/#why-not-how)

Don't put the bullets. Instead, use complete sentences.

dudect/fixture.c
```diff
@@ -74,6 +101,13 @@ static void update_statistics(const int64_t *exec_times, uint8_t *classes)
 
         /* do a t-test on the execution time */
         t_push(t, difference, classes[i]);
+
+        // t-test on cropped execution times, for several cropping thresholds.
```
jserv, line 105: Always use C-style comments.

dudect/fixture.c
```diff
@@ -56,6 +57,30 @@ static void __attribute__((noreturn)) die(void)
...
+    size_t pos = (size_t) (which * size);
+    return a_sorted[pos];
+}
+
+static int cmp(const int64_t *a, const int64_t *b)
+{
+    return *a - *b;
+}
+
+static void prepare_percentiles(int64_t *exec_times, int64_t *percentiles)
```
jserv, line 72: Show mathematical evidences here.

PR #281, opened 2025-03-20
Introduce REST API fallback

EricccTaiwan:
Switch to using the GitHub REST API as a fallback when HTML parsing fails to retrieve the commit hash from the GitHub commits page. This ensures that the script continues to work even under rate limit conditions and improves reliability. Optionally, authentication can be used to further mitigate rate limits.

Closed: #243 

Change-Id: If204a6c78bb4062bc010c3370db0ea126f99e6f5

### DIFF
--- a/scripts/check-repo.sh
+++ b/scripts/check-repo.sh
@@ -85,6 +85,28 @@ upstream_hash=$(
 
 rm -f "$temp_file"
 
+# If HTML parsing fails, fallback to using GitHub REST API
+if [ -z "$upstream_hash" ]; then
+  API_URL="https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/commits"
+  
+  # Try to use cached GitHub credentials from GitHub CLI
+  # https://docs.github.com/en/get-started/git-basics/caching-your-github-credentials-in-git
+  if command -v gh >/dev/null 2>&1; then
+    TOKEN=$(gh auth token 2>/dev/null)
+    if [ -n "$TOKEN" ]; then
+      response=$(curl -sSL -H "Authorization: token $TOKEN" "$API_URL")
+    fi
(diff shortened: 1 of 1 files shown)


scripts/check-repo.sh
```diff
@@ -85,6 +85,22 @@ upstream_hash=$(
...
 rm -f "$temp_file"
 
+# If HTML parsing fails, fallback to using GitHub REST API
+if [ -z "$upstream_hash" ]; then
+  echo "HTML retrieval failed, falling back to GitHub REST API..."
+  API_URL="https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/commits"
+  
+  # Optional: Use authentication to avoid rate limit issues.
+  # Uncomment the following line and replace YOUR_ACCESS_TOKEN with your actual token.
+  # response=$(curl -sSL -H "Authorization: token YOUR_ACCESS_TOKEN" -H "Accept: application/vnd.github+json" -H "User-Agent: YourAppName" "$API_URL")
```
jserv, line 95: Don't do that. Can you use [cached Git credentials](https://docs.github.com/en/get-started/git-basics/caching-your-github-credentials-in-git) instead?

EricccTaiwan: Ok, I'll look into it!

scripts/check-repo.sh
```diff
@@ -85,6 +85,30 @@ upstream_hash=$(
 
 rm -f "$temp_file"
 
+# If HTML parsing fails, fallback to using GitHub REST API
+if [ -z "$upstream_hash" ]; then
+  echo "HTML retrieval failed, falling back to GitHub REST API..."
```
jserv, line 90: Don't show the messages if no error happens.

PR #282, opened 2025-03-21
Correct grammar in queue.h function descriptions

salmoniscute:
This fix addresses a grammatical error in the queue.h documentation. 
When describing existence, the correct English grammar is to use "there is" rather than "there has". Therefore, I've changed "If there has only one element" to "If there is only one element" in the comments for the q_sort(), q_ascend(), q_descend(), and q_reverseK() functions.


jserv:
Thank @salmoniscute for contributing!

PR #284, opened 2025-03-27
Implement test_realloc

chensheep:
Added test_realloc implementation referencing https://danluu.com/malloc-tutorial/.

Change-Id: I87149ac233572dece7f3ec8b7d3320d07544f855

### DIFF
--- a/harness.c
+++ b/harness.c
@@ -63,6 +63,7 @@ static bool time_limited = false;
 typedef enum {
     TEST_MALLOC,
     TEST_CALLOC,
+    TEST_REALLOC,
 } alloc_t;
 
 /* Internal functions */
@@ -128,6 +129,7 @@ static void *alloc(alloc_t alloc_type, size_t size)
         char *msg_alloc_forbidden[] = {
             "Calls to malloc are disallowed",
             "Calls to calloc are disallowed",
+            "Calls to realloc are disallowed",
         };
         report_event(MSG_FATAL, "%s", msg_alloc_forbidden[alloc_type]);
         return NULL;
@@ -137,6 +139,7 @@ static void *alloc(alloc_t alloc_type, size_t size)
         char *msg_alloc_failure[] = {
             "Malloc returning NULL",
             "Calloc returning NULL",
+            "Realloc returning NULL",
         };
         report_event(MSG_WARN, "%s", msg_alloc_failure[alloc_type]);
         return NULL;
@@ -185,6 +188,28 @@ void *test_calloc(size_t nelem, size_t elsize)
     if (!nelem || !elsize || nelem > SIZE_MAX / elsize)
         return NULL;
     return alloc(TEST_CALLOC, nelem * elsize);
+}
+
+/*
+ * Implementation of adjusting the size of the memory allocated
+ * by test_malloc or test_calloc.
+ */
+void *test_realloc(void *p, size_t new_size)
+{
+    if (!p)
+        return alloc(TEST_REALLOC, new_size);
+
+    const block_element_t *b = find_header(p);
+    if (b->payload_size >= new_size)
+        return p;
+
+    void *new_ptr = alloc(TEST_REALLOC, new_size);
+    if (!new_ptr)
+        return NULL;
+    memcpy(new_ptr, p, b->payload_size);
+    test_free(p);
+
+    return new_ptr;
 }
 
 void test_free(void *p)
--- a/harness.h
+++ b/harness.h
@@ -12,9 +12,9 @@
 
 void *test_malloc(size_t size);
 void *test_calloc(size_t nmemb, size_t size);
+void *test_realloc(void *p, size_t new_size);
 void test_free(void *p);
 char *test_strdup(const char *s);
-/* FIXME: provide test_realloc as well */
 
 #ifdef INTERNAL
 
@@ -56,6 +56,7 @@ void trigger_exception(char *msg);
 /* Tested program use our versions of malloc and free */
 #define malloc test_malloc
 #define calloc test_calloc
+#define realloc test_realloc
 #define free test_free
 
 /* Use undef to avoid strdup redefined error */


harness.c
```diff
@@ -187,6 +187,33 @@ void *test_calloc(size_t nelem, size_t elsize)
...
 
+void *test_realloc(void *p, size_t size)
+{
+    /* Reference: Malloc tutorial
+     * https://danluu.com/malloc-tutorial/
+     */
+
+    if (!p) {
+        return alloc(TEST_MALLOC, size);
+    }
```
jserv, line 198: Rewrite in shorter form. i.e.,
```c
    if (!p)
        return alloc(TEST_MALLOC, size);
```

jserv: Tweak `alloc` accordingly.

harness.c
```diff
@@ -187,6 +187,33 @@ void *test_calloc(size_t nelem, size_t elsize)
...
+
+    const block_element_t *b = find_header(p);
+    if (b->payload_size >= size) {
+        return p;
+    }
```
jserv, line 203: Ditto. Use shorter form.

harness.c
```diff
@@ -187,6 +187,33 @@ void *test_calloc(size_t nelem, size_t elsize)
...
+
+    void *new_ptr;
+    new_ptr = alloc(TEST_MALLOC, size);
+    if (!new_ptr) {
+        return NULL;
+    }
```
jserv, line 209: Ditto. Use shorter form.

jserv (changes requested):
Improve git commit messages, addressing the motivation and considerations.

harness.c
```diff
@@ -187,6 +190,30 @@ void *test_calloc(size_t nelem, size_t elsize)
...
+    new_ptr = alloc(TEST_REALLOC, size);
+    if (!new_ptr)
+        return NULL;
+    memcpy(new_ptr, p, b->payload_size);
+    test_free(p);
+
+    return new_ptr;
+}
+
+
```
jserv, line 216: Remove this blank line.

harness.c
```diff
@@ -187,6 +190,30 @@ void *test_calloc(size_t nelem, size_t elsize)
     return alloc(TEST_CALLOC, nelem * elsize);
 }
 
+void *test_realloc(void *p, size_t size)
+{
+    /* Reference: Malloc tutorial
+     * https://danluu.com/malloc-tutorial/
+     */
```
jserv, line 197: Summarize instead of placing a hyperlink only.

jserv (changes requested):
Don't mention file name in the subject of git commit messages. Read https://cbea.ms/git-commit/ and `CONTRIBUTING.md` carefully. Then, improve the commit messages by writing informative sentences.

harness.c
```diff
@@ -187,6 +190,37 @@ void *test_calloc(size_t nelem, size_t elsize)
...
 }
 
+void *test_realloc(void *p, size_t new_size)
+{
+    /*
+     * If `p` is NULL, `test_realloc()` behaves like `test_malloc()`.
+     * If `p` is not NULL, check the size of the allocated memory.
+     * If `new_size` is less than or equal to the current allocation size,
+     * update the allocated memory size to `new_size`.
+     * Otherwise, allocate new memory of size `new_size`, copy the contents
```
jserv, line 200: Don't use [backtick](https://en.wikipedia.org/wiki/Backtick) characters. Check `CONTRIBUTING.md` carefully.

harness.c
```diff
@@ -187,6 +190,37 @@ void *test_calloc(size_t nelem, size_t elsize)
...
+     * from the orginal memory to the newly allocated memory,
+     * and free the original memory.
+     * Reference: Malloc tutorial
```
jserv, line 203: Drop the subject 'Malloc tutorial'.

harness.c
```diff
@@ -187,6 +190,37 @@ void *test_calloc(size_t nelem, size_t elsize)
...
+
+    if (!p)
+        return alloc(TEST_REALLOC, size);
+
+    const block_element_t *b = find_header(p);
+    if (b->payload_size >= size)
+        return p;
+
+    void *new_ptr;
+    new_ptr = alloc(TEST_REALLOC, size);
```
jserv, line 215: Merge the two lines.

harness.c
```diff
@@ -187,6 +190,41 @@ void *test_calloc(size_t nelem, size_t elsize)
...
+ *
+ * First, we check whether the memory is already allocated.
+ * If it wasn't allocated (p is NULL),
+ * the fucntion behaves like test_malloc, returning a newly allocated memory.
+ *
+ * Otherwise, we check the payload size of the orignal memory.
+ * - If new_size is less than or equal to it, return the original memory.
+ * - If new_size is greater than it, we allocate a new memory with new_size.
+ *   Copy the contents from the orginal memory to the newly allocated memory
+ *   ,and then free the original one. Finally, we return the newly one.
```
jserv, line 205: Shorten the descriptions.

harness.c
```diff
@@ -187,6 +190,41 @@ void *test_calloc(size_t nelem, size_t elsize)
...
+ * Reference: https://danluu.com/malloc-tutorial
+ */
+void *test_realloc(void *p, size_t new_size)
+{
+    if (!p)
+        return alloc(TEST_REALLOC, new_size);
+
+    const block_element_t *b = find_header(p);
+    if (b->payload_size >= new_size)
+        // TODO: Free some once we implement split.
```
jserv, line 216: Show more about this operation.

chensheep: I think the current implementation has its pros and cons. Since we do not shrink the memory size, it performs well in cases like the following:

```
malloc(100) -> realloc(1) -> realloc(2) -> realloc(3)
```

In this scenario, when we allocate a large block of memory and then reduce its size, later extensions of the memory (to a larger size) will not require additional work to handle.

Here, I propose a simple improvement:

If new_size is larger than old_size, we allocate new_size * 2 for future use.

If new_size is smaller than old_size, we check whether new_size * 2 is still smaller than old_size. If so, we allocate new memory with new_size * 2 to optimize memory usage. Otherwise, we retain the old_size.

This approach wastes some memory but reduces the number of memory copy operations when calling realloc.

Apart from the above method, I also tried implementing it using the C function realloc, but the documentation does not guarantee under what circumstances it will move the pointer (allocate new memory). This uncertainty makes it difficult to implement a noallocate_mode.

Another approach is to split the unused memory into a separate block. However, this increases the complexity of implementing test_malloc and test_free. We would need an additional field in the block to indicate whether it is free or not. In test_malloc, the simplest implementation would iterate through the list to find a free block with enough space, which increases time complexity compared to the current implementation. In test_free, to avoid excessive fragmentation, we would need to implement a function to merge adjacent free memory blocks, adding further complexity.

I am unsure whether this is worth pursuing, as this module is primarily used for detecting memory usage rather than optimizing it.

jserv: > I am unsure whether this is worth pursuing, as this module is primarily used for detecting memory usage rather than optimizing it.

We can still take this pull request without touching the potential optimizations. Refine the commits to reflect this decision.

jserv:
Thank @chensheep for contributing!

PR #285, opened 2025-03-27
Fix string operation truncation

lumynou5:
This patch fixes potential truncation caused by `strncpy`. If the destination buffer was not large enough, the string would be not null-terminated, and possibly lead to buffer overflow because of the passed buffer length depending on the source buffer.

### DIFF
--- a/web.h
+++ b/web.h
@@ -9,6 +9,6 @@ char *web_recv(int fd, struct sockaddr_in *clientaddr);
 
 void web_send(int out_fd, char *buffer);
 
-int web_eventmux(char *buf);
+int web_eventmux(char *buf, size_t buflen);
 
 #endif
--- a/linenoise.c
+++ b/linenoise.c
@@ -948,7 +948,7 @@ static int line_edit(int stdin_fd,
         char seq[5];
 
         if (eventmux_callback != NULL) {
-            int result = eventmux_callback(l.buf);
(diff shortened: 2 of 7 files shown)


web.h
```diff
@@ -9,6 +9,6 @@ char *web_recv(int fd, struct sockaddr_in *clientaddr);
 
 void web_send(int out_fd, char *buffer);
 
-int web_eventmux(char *buf);
+int web_eventmux(char *buf, size_t buflen);
```
jserv, line 12: If you want to change the function prototype, there must be a good reason.

lumynou5: Also, the compiler reports `stringop-truncation` warning for this, and stop compilation with `-O3`.

jserv (changes requested):
Make commit messages informative and reflect what you propose to change.

lumynou5:
Don't the titles describe the change? The first is an obvious mistake that the variable name is wrong, and the third fixes the buffer overflow, as the commit messages say. I suppose the reason of the overflow is also obvious because a buffer always has a finite length?
Or should I note why the function prototype has to be changed? Add a body without repetition of the title but the fact that it couldn't compile.

jserv:
> Don't the titles describe the change?

No, show the details and discuss the impact within git commit messages. Always make the development verbose and informative.

jserv:

PR #287, opened 2025-03-29
Fix memory leaks in dudect

HeatCrab:
This commit resolves memory leaks in the original implementation and improves stability with a warm-up phase. Previously, the initialization function was called multiple times within the testing function, reallocating memory for the context array without releasing earlier allocations, resulting in leaks detected by Valgrind (43,632 bytes across 909 blocks per test). Now, the initialization happens only once, and memory is released when testing completes, preventing any leaks.

Additionally, a warm-up step is added in the measurement function by discarding the first batch of data. This filters out initial anomalies (such as cache misses or memory allocation delays), aligning with the design principles of the dudect framework and enhancing measurement consistency. Previous Chinese comments are also translated to English for better readability.

These changes ensure reliability, eliminate memory issues, and adhere to the lightweight and stable testing approach outlined in the dudect paper.

Change-Id: I34b257c2ae545c7c28c0c0be60e43988efc8158d

### DIFF
--- a/dudect/fixture.c
+++ b/dudect/fixture.c
@@ -187,7 +187,7 @@ static bool doit(int mode)
     int64_t *percentiles = calloc(NUM_PERCENTILES, sizeof(int64_t));
 
     if (!before_ticks || !after_ticks || !exec_times || !classes ||
-        !input_data) {
+        !input_data || !percentiles) {
         die();
     }
 
@@ -213,8 +213,13 @@ static void init_once(void)
 {
     init_dut();
     for (size_t i = 0; i < DUDECT_TESTS; i++) {
-        ctxs[i] = malloc(sizeof(t_context_t));
-        t_init(ctxs[i]);
+        /* Check if ctxs[i] is unallocated to prevent repeated memory
+         * allocations.
+         */
+        if (!ctxs[i]) {
+            ctxs[i] = malloc(sizeof(t_context_t));
+            t_init(ctxs[i]);
+        }
     }
 }
 
@@ -222,9 +227,10 @@ static bool test_const(char *text, int mode)
 {
     bool result = false;
 
+    init_once();
+
     for (int cnt = 0; cnt < TEST_TRIES; ++cnt) {
         printf("Testing %s...(%d/%d)\n\n", text, cnt, TEST_TRIES);
-        init_once();
         for (int i = 0; i < ENOUGH_MEASURE / (N_MEASURES - DROP_SIZE * 2) + 1;
              ++i)
             result = doit(mode);
@@ -235,6 +241,7 @@ static bool test_const(char *text, int mode)
 
     for (size_t i = 0; i < DUDECT_TESTS; i++) {
         free(ctxs[i]);
+        ctxs[i] = NULL;
     }
 
     return result;


jserv (changes requested):
Refine the git commit messages. In particular, the statement "Previous Chinese comments are also translated to English for better readability" should not appear since there were no Chinese characters appearing in the git log.

Also, don't try to summarize as following:
> These changes ensure reliability, eliminate memory issues, and adhere to the lightweight and stable testing approach outlined in the dudect paper.

Write down your thoughts as a human being.

HeatCrab:
Received. I’ve made the adjustments. The comment about Chinese was due to
using Chinese in my own branch, and I’ll be mindful of this in the future.
I’ve also removed the overly stiff and unnecessary summary, adding my own
thoughts and reasons for the code changes instead. Please review again.

> new commit [8677947](https://github.com/sysprog21/lab0-c/commit/86779472d8cb117bb5f892051fb2a218e8f46df5)

jserv (changes requested):
Avoid bullets in git commit messages. Instead, use **plain and complete English sentences**. You should discuss the root cause resulting in memory leaks and why you coupled the fix for memory leaks with the warm-up procedure for duduct in the proposed pull request.

jserv:
> I’ve submitted a new commit with these changes—happy to adjust if you’d prefer them separated.

Check `CONTRIBUTING.md` carefully:
> Group Related Changes Together: Each commit should encapsulate a single, coherent change. e.g., if you are addressing two separate bugs, create two distinct commits. This approach produces focused, small commits that simplify understanding, enable quick rollbacks, and foster efficient peer reviews. By taking advantage of Git’s staging area and selective file staging, you can craft granular commits that make collaboration smoother and more transparent.

You should provide at least two pull requests to address the proposed changes. That is, this pull request should be refined to resolve memory leaks found in 'dudect' directory, and the incoming pull request will deal with the warm-up step.

jserv (changes requested):
Don't mention the file name in the subject of git commit messages. Instead, refer to its module name such as 'dudect'.

HeatCrab:
OK! Updated and adjusted the title name.

> latest commit [350b6fa](https://github.com/sysprog21/lab0-c/commit/350b6fae18742cdd728dd652cad5a7ab3816a41a)

HeatCrab:
OK! Updated and adjusted the subject of git commit messages.

> latest commit [e1216bb](https://github.com/sysprog21/labx0-c/commit/e1216bb21f2ec3d9a2ac23ac8dde46d81ce3a327)

dudect/fixture.c
```diff
@@ -209,22 +220,14 @@ static bool doit(int mode)
...
-        ctxs[i] = malloc(sizeof(t_context_t));
-        t_init(ctxs[i]);
-    }
-}
-
 static bool test_const(char *text, int mode)
 {
     bool result = false;
 
+    init_once();  // Initialize only once
```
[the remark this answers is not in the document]
HeatCrab: After thinking it over, I realized the function name already conveys the
comment’s meaning, and I couldn’t come up with a more concise or clear
alternative, so I decided to remove it.

> latest commit [7709ce1](https://github.com/sysprog21/lab0-c/commit/7709ce1251bb1a9975487acc3d54d26aece7f504)

dudect/fixture.c
```diff
@@ -213,18 +213,21 @@ static void init_once(void)
 {
     init_dut();
     for (size_t i = 0; i < DUDECT_TESTS; i++) {
-        ctxs[i] = malloc(sizeof(t_context_t));
-        t_init(ctxs[i]);
+        if (!ctxs[i]) {  // Avoid repeated allocation
```
jserv, line 216: Use C-style comments.

HeatCrab: Fixed !

> new commit [00cd93a](https://github.com/sysprog21/lab0-c/commit/00cd93ac61096eda8eaed75628fc469b20d980e5)

jserv:
Thank @HeatCrab for contributing!

Issue #265, opened 2025-03-13
`fmtscan` reports unique bad spellings but does not specify where


EricccTaiwan:
When committing , `fmtscan` reports "3 unique bad spellings found (3 non-unique)" but does not indicate which words are misspelled or their locations in the source code.

It would be helpful if fmtscan can automatically displayed more details about spelling errors.

```shell
$ git commit -a
Following files were changed:
  - queue.c : 8 insertions(+), 7 deletions(-)
Running fmtscan...
href
monospace
rel
7804 lines scanned (0.237M bytes)
53 printf style statements being processed
3 unique bad spellings found (3 non-unique)

[!] Check format strings for spelling
$
```

Urbaner3:
Better specify when you commit by providing messages generated by `tig` or `git reflog`. I can't be sure what is in stashes and local changes. 
I see that you use printf related functions 53 times and 3 of them went wrong. There is too little information to check the context. 
BTW, check #274 to see my error, and it's also about fmtscan.

Urbaner3:
```shell=
(base) urbaner@urbaner-wtumagic:~/linux2025/lab0-c$ git commit
Following files were changed:
  - dudect/fixture.c : 96 insertions(+), 11 deletions(-)
Running fmtscan...
meas
8119 lines scanned (0.253M bytes)
53 printf style statements being processed
1 unique bad spellings found (1 non-unique)

[!] Check format strings for spelling
(base) urbaner@urbaner-wtumagic:~/linux2025/lab0-c$ git rst --staged dudect/fixture.c
(base) urbaner@urbaner-wtumagic:~/linux2025/lab0-c$ git diff
diff --git a/dudect/fixture.c b/dudect/fixture.c
index fbf4eb4..dab3934 100644
--- a/dudect/fixture.c
+++ b/dudect/fixture.c
```
I got the report. There is an output message.
>Running fmtscan...
meas

Search through the `git diff` records, I found exact the same syntax "meas" in printf. Yes, it's exactly in printf. 
```diff
+    double number_traces_max_t = tmp->n[0] + tmp->n[1];
     double max_tau = max_t / sqrt(number_traces_max_t);
 
     printf("\033[A\033[2K");
-    printf("measure: %7.2lf M, ", (number_traces_max_t / 1e6));
+    printf("meas: %7.2lf M, ", (number_traces_max_t / 1e6));
     if (number_traces_max_t < ENOUGH_MEASURE) {
[...]

PR #288, opened 2025-04-04
Enhance `fmtscan` dictionary with HTML and CSS terms

HeatCrab:
This PR updates the spelling dictionary with three terms - 'href' and 'rel' from HTML, and 'monospace' from CSS - that fmtscan marked as false positives during web integration. It fixes a flagging issue also seen in #265, ensuring reliable static analysis, but leaves #265's request for better error details in fmtscan unaddressed.

Related Issue: #265

### DIFF
--- a/scripts/aspell-pws
+++ b/scripts/aspell-pws
@@ -143,6 +143,7 @@ github
 glibc
 gprof
 hotfix
+href
 http
 https
 hv
@@ -201,6 +202,7 @@ mknod
 mlock
 mlockall
 mmap
+monospace
 mprotect
 mq
 mrs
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
The passage effectively cautions against using the ambiguous term "front-end" by providing several distinct technical contexts where it has different meanings. It then offers clearer alternative terminology ("web" or "http").

HeatCrab:
I've updated both the PR title and description to use "HTML and CSS terms" instead, specifying that 'href' and 'rel' come from HTML, and 'monospace' from CSS. Please let me know if anything else needs improvement!

jserv:
Thank @HeatCrab for contributing!

PR #293, opened 2025-04-17
Allow inline assembly

lumynou5:
This patch is an attempt to enable inline assembly not being blocked by the checks.

Previously, fmtscan would report spelling in inline assembly and block committing, so an additional check is added to skip identifiers that are `asm` and contents in a pair of parentheses following them.

Also, Cppcheck cannot recognize the GNU extension of explicit register variables:
```
register int x asm("eax");
```

This patch suppresses syntax error to prevent explicit register variables being blocked. It also makes real syntax error suppressed, but it doesn't affect much while real ones lead to compile-time errors and can be found early.

### DIFF
--- a/scripts/pre-commit.hook
+++ b/scripts/pre-commit.hook
@@ -26,6 +26,7 @@ cppcheck_suppressions() {
     "missingIncludeSystem"
     "noValidConfiguration"
     "unusedFunction"
+    "syntaxError"
     "identicalInnerCondition:log2_lshift16.h"
     "checkLevelNormal:log2_lshift16.h"
     "nullPointerRedundantCheck:report.c"


jserv:
Could your code handle all variations of inline assembly syntax found in GNU-compatible compilers like Clang, including `asm`, `__asm`, and `__asm__`?

lumynou5:
> Could your code handle all variations of inline assembly syntax found in GNU-compatible compilers like Clang, including `asm`, `__asm`, and `__asm__`?

It can handle `__asm__` now.

I didn't find any document of `__asm` except the very different syntax of MSVC.

jserv:
> I didn't find any document of `__asm` except the very different syntax of MSVC.

Simply search `__asm` inside [GCC repository](https://github.com/gcc-mirror/gcc). It was a historical keyword.

jserv (changes requested):
Squash commits for reviewing.

jserv:

